PR #2, opened 2024-02-24
fix(client): add missing localization files; cleanup;

lmssiehdev:


leia-uwu:
uh the localization files contain post kong stuff
I wonder if the pre kong ones were archived anywhere

lmssiehdev:
> I wonder if the pre-kong ones were archived anywhere

I found some but they're missing a lot of data and the translation for some of the fields got updated since then.

> or we can just make a script to remove the unused fields and only keep the ones used by the post kong client lol

I can work on this, should I discard any of the keys not in english.js?

leia-uwu:
I already did it before merging lol

lmssiehdev:
noice, not sure how I missed that.

PR #7, opened 2024-03-01
refactor(client): biggest one yet

lmssiehdev:
amviance.js 😭

### DIFF
--- a/client/src/game.js
+++ b/client/src/game.js
@@ -785,7 +785,7 @@ export class Game {
                 this.U = true;
             }
             if (a && this.ur) {
-                helpers.U(this);
+                helpers.cheatDetected(this);
             }
         }
     }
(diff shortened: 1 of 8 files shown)


client/src/helpers.js
```diff
@@ -49,32 +51,9 @@ const helpers = {
         }
         e.appendChild(o[0]);
     },
-    Z: function() {
```
lmssiehdev, line 52: this was used to break malicious proxies that remove the notification modal from the html

client/src/game.js
```diff
@@ -283,306 +294,337 @@ export class Game {
 
     warnPageReload() {
         return (
+            import.meta.env.PROD &&
```
lmssiehdev, line 297: disable the alert when reloading during development, hate closing it every time I save.

PR #11, opened 2024-03-07
feat: add emotes support

lmssiehdev:
gib feedback

### DIFF
--- a/server/src/game.ts
+++ b/server/src/game.ts
@@ -18,13 +18,13 @@ import { EmotesDefs } from "../../shared/defs/gameObjects/emoteDefs";
 
 export class Emote {
     playerId: number;
-    position: Vec2;
+    pos: Vec2;
     type: string;
     isPing: boolean;
 
-    constructor(playerId: number, position: Vec2, type: string, isPing: boolean) {
+    constructor(playerId: number, pos: Vec2, type: string, isPing: boolean) {
         this.playerId = playerId;
-        this.position = position;
+        this.pos = pos;
         this.type = type;
         this.isPing = isPing;
     }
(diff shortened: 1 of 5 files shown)


shared/net.js
```diff
@@ -1140,7 +1155,7 @@ class EmoteMsg {
         this.pos = s.readVec(0, 0, 1024, 1024, 16);
         this.type = s.readGameType();
         this.isPing = s.readBoolean();
-        s.readBits(6);
```
lmssiehdev, line 1143: couldn't read 6 bits

server/src/game.ts
```diff
@@ -107,8 +123,17 @@ export class Game {
...
+                for (const emote of this.emotes) {
+                    if (emote.playerId === player.id || !emote.isPing) {
+                        player.emotes.add(emote);
+                    }
+                }
+            }
+
             player.sendMsgs();
         }
+        if (this.emotes.size) this.emotes = new Set<Emote>();
```
leia-uwu: just do `this.emotes.clear()`

shared/net.js
```diff
@@ -904,7 +908,13 @@ class JoinMsg {
         this.proxy = s.readBoolean();
         this.otherProxy = s.readBoolean();
         this.bot = s.readBoolean();
-        s.readAlignToNextByte();
```
leia-uwu, line 907: should keep the byte alignment

shared/net.js
leia-uwu: because its 5 not 6, oops, my fault

leia-uwu (changes requested):

shared/net.js
lmssiehdev: oops, didn't intend to delete it.

PR #20, opened 2024-03-24
fixed equipothergun input action

MamounKolovos:


server/src/objects/player.ts
```diff
@@ -909,13 +912,23 @@ export class Player extends BaseGameObject {
...
+                //             this.weapons[i] !== this.weapons[this.curWeapIdx]) {
+                //         this.curWeapIdx = i;
+                //         break;
+                //     }
+                // }
+
+                // completely unreadable but optimized weapon swapping code since leia apparently values fewer lines of code over readability
+                if (this.curWeapIdx == 0 || this.curWeapIdx == 1) {
+                    const otherGunSlotIdx = this.curWeapIdx ^ 1;
+                    const isOtherGunSlotFull: number = +!!this.weapons[otherGunSlotIdx].type;//! ! converts string to boolean, + coerces boolean to number
```
lmssiehdev, line 924: +!! 💀 

I want to join the refactor train, here is my attempt.

```js
case GameConfig.Input.EquipOtherGun:
    if (this.weapons[0]?.type && this.weapons[1]?.type) this.curWeapIdx = this.curWeapIdx === 0 ? 1 : 0;
    else {
        if ( this.curWeapIdx >= 2 && this.weapons[0].type ) this.curWeapIdx = 0; 
        else if ( this.curWeapIdx >= 2 && this.weapons[1].type ) this.curWeapIdx = 1; 
        else this.curWeapIdx = 2;
    }
break;
```

MamounKolovos: This isn't correct, equipothergun doesn't only work if you have a gun selected and switch to the other gun slot. It also deals with melee which is why my original code is necessarily more complex than your re-factor.

lmssiehdev: damn, I really liked that two-line solution, it should handle switching from throwables too? I gave it another try and edited my solution, please take a look.

MamounKolovos: my solution is how original surviv's equipothergun input action worked, it never worked with throwables. I'm 99.9% sure my original code is the most efficient way to program this.

lmssiehdev: maybe it's the most efficient, I can't judge that. but it is unreadable and difficult to parse, at least for me.

> it never worked with throwables.

interesting, but doing nothing when the throwable slot is selected doesn't make sense.

MamounKolovos: > but it is unreadable and difficult to parse

yeah i did it like that intentionally since leia wants everything to be ultra optimized and values less lines over readability. My original solution worked the same with more if statements and was much clearer to understand and she deleted it.

> but doing nothing when the throwable slot is selected doesn't make sense

I agree, I asked people in the community whats supposed to happen in original surviv when you run the input with a throwable selected.

leia-uwu:
pls remove debugging code smh (like adding guns to player inventory)

PR #21, opened 2024-03-24
fix(server): enable gun pickup with empty slot

lmssiehdev:


leia-uwu:
me when I'm refactoring whole loot pickup code
will still merge because of the client fixes

lmssiehdev:
lol sorry, I could have reverted those before you merged.

PR #22, opened 2024-03-24
feat(server): add support for equipping next/prev scope

lmssiehdev:
High priority for me, low priority for thee


lmssiehdev:
I only pushed a few chunks of player.ts so I can't get the ci lint to pass, feel free to merge if the changes look good.

leia-uwu:
from looking at the code, I don' think it will work if you don't have all scopes since it just checks for previous one
like it won't switch from a 8x scope to a 2x scope

lmssiehdev:
sometimes I forget players spawn with an empty inventory 😅

PR #26, opened 2024-04-06
feat(server): redzone

lmssiehdev:


server/src/objects/gas.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    countdownStart = 0;
+
+    /** Gas Timer */
+    gasT = 0;
+
+    dirty = true;
+    timeDirty = true;
+
+    get data(): GasData {
```
leia-uwu: why this getter? just set the packet gasData to the class instance

server/src/objects/gas.ts
```diff
@@ -0,0 +1,257 @@
...
+        return math.distanceBetween(pos, this.currentPos) >= this.currentRad;
+    }
+}
+
+const GasStages = [
+    {
+        mode: GasMode.Inactive,
+        duration: 0,
+        radOld: 534.6,
+        radNew: 534.6,
```
leia-uwu: gas radius should be based on map size (a value between 0 and 1) not hardcoded values

shared/utils/math.ts
```diff
@@ -10,6 +10,15 @@ export const math = {
         return a < max ? a > min ? a : min : max;
     },
 
+    /**
+     * Generate a random floating-point value.
+     * @param min The minimum value that can be generated.
+     * @param max The maximum value that can be generated.
+    */
+    randomFloat(min: number, max: number): number {
```
leia-uwu: we already have a random float function on util.ts
https://github.com/leia-uwu/resurviv/blob/master/shared/utils/util.ts#L36

leia-uwu (changes requested):

server/src/objects/gas.ts
```diff
@@ -138,119 +127,119 @@ const GasStages = [
     {
         mode: GasMode.Inactive,
         duration: 0,
-        radOld: 534.6,
-        radNew: 534.6,
+        radOld: 0.7,
```
lmssiehdev, line 130: 0.64 ( value in the wiki ) was showing the red zone even when it's inactive

shared/utils/math.ts
```diff
@@ -231,6 +232,31 @@ export const math = {
         return inside;
     },
 
+    /**
+     * Calculate the distance between two points.
+     * @param v1 The first point.
+     * @param v2 The second point.
+     */
+    distanceBetween(v1: Vec2, v2: Vec2): number {
```
leia-uwu: theres `v2.distance`

PR #27, opened 2024-04-13
feat(server): loadout

lmssiehdev:
give feedback


shared/net.js
```diff
@@ -910,12 +920,17 @@ export class JoinMsg {
         this.useTouch = s.readBoolean();
         this.isMobile = s.readBoolean();
         this.bot = s.readBoolean();
-        this.emotes = [];
+
+        this.loadout.outfit = s.readString();
+        this.loadout.melee = s.readString();
+        this.loadout.heal = s.readString();
+        this.loadout.boost = s.readString();
```
leia-uwu: use write/readGameType

leia-uwu (changes requested):

shared/defs/gameObjects/unlockDefs.js
```diff
@@ -1,8 +1,45 @@
...
+/**
+ * Checks if an item is present in the player's loadout
+ * @param {string} item
+ * @param {"boost" | "heal" | "melee" | "emote" | "outfit"} category
+ * @returns { boolean }
+ */
+export function isItemInLoadout(item, category) {
+    switch (category) {
+    case "outfit": {
+        return item in OutfitDefs;
```
leia-uwu: uh should probably check if the outfit is in the unlock defs otherwise players can send Ghillie suit as their skin and it will accept it

leia-uwu (changes requested):

PR #31, opened 2024-05-03
Fixed gun picking and reloading issues

GatoVuelta:
- Fixed a weird bug that prevented the player from picking a gun (or replacing the main one) under certain conditions, for example, having a gun that is able to become dual and picking another gun while the first is on hand, you can test by trying to pick up a MP220 while holding an M9.
- Fixed a bug that interrupted the gun reloading if you swapped slots.

(sory for bad england)

### DIFF
--- a/shared/gameConfig.ts
+++ b/shared/gameConfig.ts
@@ -55,7 +55,6 @@ export const GameConfig = {
         Throwable: 3,
         Count: 4
     },
-    GunSlots: [0, 1],
     WeaponType: ["gun", "gun", "melee", "throwable"],
     DamageType: {
         Player: 0,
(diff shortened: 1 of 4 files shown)


GatoVuelta:
@leia-uwu please review it

shared/gameConfig.ts
```diff
@@ -55,6 +55,7 @@ export const GameConfig = {
         Throwable: 3,
         Count: 4
     },
+    GunSlots: [0, 1],
```
leia-uwu: just use the WeaponType array instead of adding a new field to gameConfig

leia-uwu (changes requested):

PR #32, opened 2024-05-09
fixed auto reload not working when swap slots and after popping adren…

MamounKolovos:
…/heals


leia-uwu:
can you like remove the debug code (loot spawning and player inventory)

PR #35, opened 2024-05-24
Fixed gas not doing any damage

GatoVuelta:


server/src/game.ts
```diff
@@ -81,7 +81,10 @@ export class Game {
         this.id = id;
         this.logger = new Logger(`Game #${this.id}`);
         this.logger.log("Creating");
-        const start = Date.now();
+
+        const now = Date.now();
+        const start = now;
+        this.now = now;
```
leia-uwu, line 87: this is wrong and will cause the first tick delta time to be way too high

leia-uwu (changes requested):

leia-uwu:
fixed in 11c18a248cde02cea820d92342407765bc0aba8c

PR #38, opened 2024-06-08
fix up gun play stuff, mostly qs

ei-pi:
at least carbon-copy stuff _correctly_ lol
have fun
read second commit's desc btw


server/src/objects/player.ts
```diff
@@ -356,6 +358,10 @@ export class Player extends BaseGameObject {
         }
         this.inventory["1xscope"] = 1;
         this.inventory[this.scope] = 1;
+
+        // (async() => {
+        //     (await import("./qsDemos")).getDemos(this)["two-shot overclock"]();
```
leia-uwu: can you remove the commented out code and the qsDemos.ts file

leia-uwu (changes requested):

server/src/objects/player.ts
ei-pi: could've diy after-the-fact but whatever
commit is there for anyone to checkout if they want to

a-beyblade:
@ei-pi surviv-underclock/docs does not make the claim below. It defines free switches as switches that occur at least 1 second after the last free switch, without any reference to the fire timer, fire delay, or ability for the weapon to fire.

> the paper claims that if the weapon you're switching from is currently unable to fire, then a switch cannot be free—however this ruleset directly conflicts with some sequences later presented (which have been done ingame); in reality, the old weapon's cooldown timer is completely irrelevant

There is an issue with the Math section that incorrectly specifies free switches that occur exactly 1 second apart, which has just been fixed in surviv-underclock/docs@bbe77dc021c7787ff86ea331ef515c84a6e6a8bd.

ei-pi:
now I realize I read it wrong, thanks

a-beyblade:
@ei-pi I will add that the fire timer is relevant for effective switch delay (but not free switch) when the deploy groups are identical, which your code does not handle correctly. For example, if you have M870 equipped and free switch to M870, the effective switch delay is 250 ms if and only if the first M870 had an expired fire timer.

This part

```ts
swappingToGun && curWeaponDef.deployGroup / nextWeaponDef.deployGroup === 1
```

should be

```ts
swappingToGun && curWeaponDef.deployGroup / nextWeaponDef.deployGroup === 1 && curWeapon.cooldown > this.player.game.now
```

Based on your code, I probably would implement it as

```ts

const isFreeSwitch = this.player.freeSwitchTimer <= this.player.game.now

if (isFreeSwitch) {
	this.player.freeSwitchTimer = this.player.game.now + (GameConfig.player.freeSwitchCooldown * 1000);
}

const swappingToGun = nextWeaponDef.type == "gun";
const effectiveSwitchDelay = swappingToGun
	? isFreeSwitch && !(curWeaponDef.deployGroup && curWeaponDef.deployGroup === nextWeaponDef.deployGroup && curWeapon.cooldown > this.player.game.now)
		? GameConfig.player.baseSwitchDelay
		: nextWeaponDef.switchDelay
	: 0;

nextWeapon.cooldown = this.player.game.now + (effectiveSwitchDelay * 1000);
```

This condition
```ts
!(curWeaponDef.deployGroup && curWeaponDef.deployGroup === nextWeaponDef.deployGroup && curWeapon.cooldown > this.player.game.now)
```

could also be written as

```ts
(!curWeaponDef.deployGroup || curWeaponDef.deployGroup !== nextWeaponDef.deployGroup || curWeapon.cooldown <= this.player.game.now)
```

ei-pi:
I was gonna fix it but then didn't get around to it lol

leia-uwu:
fixed on 462f0e8

a-beyblade:
To nitpick, `this.player.freeSwitchTimer < 0` should be `this.player.freeSwitchTimer <= 0`, but I don't think it matters enough to really care.

leia-uwu:
the chance of it being exactly 0 is... near 0 (haha) due to how the timer works lol

PR #39, opened 2024-06-09
Feat/team menu

MamounKolovos:


client/src/objects/player.js
leia-uwu, line 1: why this change

client/src/game.js
```diff
@@ -967,6 +967,7 @@ export class Game {
         }
         // Update player status
         if (msg.playerStatusDirty) {
+            // console.log("hello");
```
leia-uwu: :p remove this pls

shared/gameConfig.ts
leia-uwu, line 1: should be on server config since its only used by the server and not by the client

client/src/objects/player.js
MamounKolovos: player indicators on the minimap fade if i keep it uncommented

shared/gameConfig.ts
lmssiehdev: I will be using it here as `[TeamMode.Solo]: "solo"`

https://github.com/leia-uwu/resurviv/blob/0a85a99759b0d4c7ea5bbab651660ecb3b4bd557/client/src/siteInfo.js#L30-L35

client/src/objects/player.js
leia-uwu: i think theres something wrong with the server code then because that code is from the original client, they shouldn't fade

PR #46, opened 2024-06-28
added plugin manager + merge

MamounKolovos:


server/src/plugins/survivReloadedPlugin.ts
leia-uwu, line 1: uh i don't want this plugin upstream

leia-uwu (changes requested):

PR #47, opened 2024-06-28
refactor: clean up client HTML; fix tiny ts errors;

lmssiehdev:


client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "build": "vite build",
+    "build": "tsc && vite build",
```
leia-uwu, line 4: why this

leia-uwu (changes requested):
i would like to keep PWA metadata so people can install it from the browser on phones

client/package.json
lmssiehdev: Vite does not perform type checking

leia-uwu: hmm change it to `tsc --noEmit` so it doesn't output the files 
but maybe we should just use this https://github.com/fi3ework/vite-plugin-checker to get TS errors in the browser on dev server

leia-uwu:
btw is the pwd generator dep really necessary? it requires libvips to be installed on linux otherwise `pnpm install` will fail

leia-uwu:
tbh it seems more like a tool you would install globally on your system to use the cli, i will remove it from the dependency for now but keep the package.json script

leia-uwu:
:stuck_out_tongue: i don't have push permission to your repo so i'm going to merge and do my changes after

PR #54, opened 2024-07-02
Update HOSTING.md

chpsterz:
Update the guide so that the client can actually connect to the server since that auto system is not working atm


leia-uwu:
i would rather fix the system than do this hack

PR #60, opened 2024-07-05
can't join game after grace period ends

MamounKolovos:


MamounKolovos:
fixed being able to spectate self
fixed downed players being able to shoot
fixed not being able to revive if next to loot and refactored interaction code
added custom movement speed object to gameconfig (undefined by default)
fixed being able to get boost/adrenaline while knocked

PR #68, opened 2024-07-10
feat: add bots again;

lmssiehdev:
bots are disabled by default, add /?bots=true query to enable them.

now with normal commits


blankdrive:
how to activate bots?

PR #74, opened 2024-07-14
better nade throwing distance estimate

MamounKolovos:


server/src/game/objects/obstacle.ts
```diff
@@ -307,6 +307,19 @@ export class Obstacle extends BaseGameObject {
             v2.set(lootPos, v2.add(this.pos, v2.rotate(def.lootSpawn.offset, this.rot)));
         }
 
+        if (
+            params.source instanceof Player &&
+            params.source.hasPerk("scavenger") &&
+            def.loot.length != 0
```
leia-uwu: if player had scavenger obstacles that have loot by default would still drop an extra item

server/src/game/objects/player.ts
```diff
@@ -1303,6 +1303,8 @@ export class Player extends BaseGameObject {
             }
         }
 
+        if (this.hasPerk("bonus_assault")) finalDamage *= 1.08;
```
leia-uwu: pretty sure that only applied to bullets and should be done in bullet.ts by checking if the trail is saturated

leia-uwu (changes requested):

PR #82, opened 2024-07-17
reverted splinter and fix damageMult perks logic

MamounKolovos:


shared/utils/util.ts
```diff
@@ -63,6 +63,11 @@ export const util = {
         return Math.floor(Math.random() * (max - min + 1)) + min;
     },
 
+    randomPointOnCircle(rad: number) {
```
leia-uwu, line 66: why this function instead of `util.randomPointInCircle`

leia-uwu (changes requested):

PR #86, opened 2024-07-19
added minimum spawn radius beetween same tick non-flare airdrops

b4a3fc:


leia-uwu:
closing this as it has a lot of unrelated changes and i don't think this airdrop minimum spawn distance existed in the original game since on 50v50 it was common to see air drops spawn really close at the end zone stages

PR #95, opened 2024-08-20
Add missing files, change logo to survev

IOMAN12:


lmssiehdev:
can't you just remove the reference to the files inside index.html instead of adding them?

PR #98, opened 2024-08-20
add bleed damage multiplier

MamounKolovos:


server/src/game/objects/player.ts
```diff
@@ -1780,6 +1792,10 @@ export class Player extends BaseGameObject {
         this.shootHold = false;
         this.cancelAction();
 
+        if (this.game.map.mapDef.gameConfig.bleedDamageMult != 1) {
+            this.bleedDamage *= this.game.map.mapDef.gameConfig.bleedDamageMult;
```
leia-uwu: this will multiply the bleed damage even on first time the player is downed
i think it should only multiply it on next times
i would refactor player `bleedDamage` to be `timesDowned = 0` that would be incremented every time the player got downed and then bleed damage would be `mapDef.gameConfig.bleedDamage * (this.timesDowned * mapDef.gameCOnfig.bleedDamageMult)`

PR #99, opened 2024-08-21
fix few css bugs; minor client refactoring

lmssiehdev:


client/src/api.ts
leia-uwu, line 1: i would like to keep api.ts because it works like a config for the api URL

client/l10n/en.json
leia-uwu, line 1: i don't think some translations keys from here should have been deleted
like `index-google` (when we add accounts i want google and discord login support, don't care about others tho),

leia-uwu:
also idk about binding equip other gun to something by default
i'm a fan of that but don't want to keep adding exceptions to 1:1 remake xd

lmssiehdev:
Both the "Equip Other Gun" and space weren’t set to anything, so it’s not really an issue, and most players set it themselves. I think it’s still worth making it the default.

I won’t revert it, but feel free to do so.

PR #115, opened 2024-09-07
Change title img to survev and other stuff

IOMAN12:


lmssiehdev:
what's the reason for removing the pwa and local fonts changes?

lmssiehdev:
I see. I only added the Latin font variant and it looks like it doesn’t support Vietnamese and Japanese languages. will fix later.

PR #117, opened 2024-09-08
refactor: remove leftover pwa code

lmssiehdev:
I'm still not sure why pwa code was removed, there is no alternative yet and I and others still use it with the offline version.


leia-uwu:
wdym no alternative? the manifest.json works just fine for me

lmssiehdev:
but that doesn't cache assets right? you still have to be connected to the internet to play.

leia-uwu:
yeah it apparently doesn't
maybe we can keep it on offline branch only?

lmssiehdev:
sure, that will do.

PR #125, opened 2024-09-29
Fixed damageTaken stat displaying a 99 occasionally

TheHaloDeveloper:
Math.max() picks the of greater value of the 2 provided, which means if the damage taken is less than 100, it will return 100. On the other hand, if the player heals and has more than 100 damage taken, then that value will be returned.


TheHaloDeveloper:
oops

TheHaloDeveloper:
Screenshot 2024-09-29 at 3 58 31 PM

example of what i mean btw

leia-uwu:
this should be done in the server not client
and also your fix makes it so if you win while taking less than 100 damage it will show up as 100

PR #131, opened 2024-10-15
fiox the gaem and maek it beirtyer

DamienVesper:
i amdem hte game beter


leia-uwu (approved):
Thanks!
don't know how we lived with that game breaking exploit for so long

DamienVesper:
pls merge !!!

PR #135, opened 2025-02-26
Detection for krity

nostopgmaming17:
Krity asset swaps game and primarily the init method
image


Drino955:
LOLLLL, pointless, noob

PR #142, opened 2025-03-13
Atlas generation

leia-uwu:
For now i'm going to commit the generated atlases mainly because:

Some SVGs break with node-canvas so when building i first export everything to PNGs with inkscape
(see the export.sh script)

And because it would make the dev server, CI builds, etc slow as shit like the suroi ones... i have had enough of that
Also some people hosting survev literally dont have enough ram to build the atlases on their servers 😭


leia-uwu:
also while the packing algorithm goes either vertical or horizontally unlike original surviv that tried to stay square
it still produces the same amount of spritesheets so no space is wasted

PR #152, opened 2025-03-21
Cobalt Loot Tables/Savannah Map Work

NotASwifty13:
Savannah Map needs a little touch up before merging, mostly on the lakes. Opinions on map size, spawns, and loot tables welcome before merging. Loot table weights are sorta an attempt to get kinda close to the original, but are mostly just placeholders for when IOMAN comes through and fixes everything. 

Same goes for the cobalt tables.


shared/defs/maps/savannahDefs.ts
```diff
@@ -27,6 +30,214 @@ const mapDef: PartialMapDef = {
...
+        tier_airdrop_ammo: [
+            { name: "9mm", count: 30, weight: 3 },
+            { name: "556mm", count: 30, weight: 3 },
+            { name: "762mm", count: 30, weight: 3 },
+            { name: "45acp", count: 30, weight: 3 },
+        ],
+    },
+    mapGen: {
+        map: {
+            baseWidth: 565,
```
leia-uwu: should keep same baseWidth and baseHeight as main mode instead of overwriting them, and only overwrite the shore and grass inset, all other modes do that

PR #153, opened 2025-03-21
Accounts

leia-uwu:
Todo / stuff to fix:

- [x] Fix the duplicated players and game saving logic
- [x] Make the game server request API to save the game since they wont be able to access the DB if they are on different regions (assigned: leia)
- [x] ~~Leaderboard most kills sorting seems to be broken, i think i may have broken it when porting to PostgreSQL? nvm it works now lol~~
- [x] Refactor team menu and a bit of the remains of the old API server because they are a mess (assigned: leia)
- [x] Make /stats:slug work on production (assigned: leia)
- [x] Share more API request body and response typings between client and server (assigned: lmssiehdev)
- [x] Finish IP bans (assigned: lmssiehdev)
- [x] Log IP's on save_game now that the game server doesn't have access to the database anymore
- ~~[ ] Port the captcha from prod but in a way less stupid and annoying way~~ will do after the merge
- [x] Log server errors to a discord webhook too (also need to add config options for that, rn theres just a hardcoded earlier return to disable it) or maybe log them to the database (though that would require like a proper dashboard and stuff ig to visualize them easily)
- [x] actually try and fix the leaderboard caching (assigned: lmssiehdev) 
- [x] Match surviv behavior (I didn't do these either because I didn't know the original behavior or it was too bothersome that I just gave up)
    - ~~[x] find a way to make /stats:slug work in a non sucky way with vite~~
    - ~~[x] fix the slug and username being the same~~
    - [x] fix /api/match_history getting all-time match data 
- [ ] 50v50 mode? i have no clue how the stats looked for that.. HELP!!!
    
will add more as i remember / notice them

btw @lmssiehdev so we don't step on each other toes feel free to edit this description to add what you are working on atm by adding an assigned parenthesis ig, and also add new items


leia-uwu:
uh Lucia auth was apparently depreacted https://github.com/lucia-auth/lucia/discussions/1714

should we bother switching or writing our own auth?

leia-uwu:
unrelated to lucia itself but the game process wont be able to open the database to verify sessions on prod (since api and database will only live on e.g NA, other regions wont have direct access to the DB)

so the find_game auth flow will require a bit of a refactor
my plan is to have it like this:

the api server can get the cookies on both places find game is called, /api/find_game and /team_menu
we use that to send to the game server the account information on the game server /find_game
so the game server /find_game will accept an array of account slugs (and maybe more account data it needs?) and returns join tokens associated with each slug
(array because team menu has to request it for multiple players)
then we send each join token to their respective players on team_menu etc

and for each team menu player we send the join token associated with them

thoughts?

leia-uwu:
also removing all database imports from the game process makes the game creation startup time go from 2 seconds to 600ms on the dev server lol
(the full time from the process being spawned to the game fully generating, the "created in" log we have rn is kinda a lie since it doesn't count process creation time lol)

leia-uwu:
I have deployed a test server for accounts on https://survev.leia-is.gay/

only did discord oauth for it because i was lazy to setup google stuff
but its a full production build working!

lmssiehdev:
Damn, I was away just for a couple of days, and you have already solved most of the issues, even the ones I didn't get the chance to document yet. :sob:

For error logs, honestly, if we require any more data than what is received from the webhook, we should just use Sentry or PostHog or other proper error tracking tools. I'm not a fan of maintaining a separate dashboard just for this.

as for Lucia.. there is no better alternative right now, and the migration from v3 seems doable. I'll give it a try.

and ehhh how should we handle this?

https://survev.leia-is.gay/stats/preacher

leia-uwu:
> for error logs...

yeah tbh lets just set up a proper error tracking tool idk what i was on when i said a dashboard just for that

> and ehhh how should we handle this?

you mean players joining the same game twice? i discussed this on discord and we decided its fine..
the UI breaking and the dropdown opening on the first game instead of the one you clicked is annoying tho, maybe we can just filter out duplicated game IDs from the match history route?
and could also invalidate other players that join with the same account id while theres one thats already alive
so you can still rejoin the same game after dying but cant join it with 2 alive players, idk

lmssiehdev:
> and could also invalidate other players ...

players inflating their game would be a real issue especially in slow-filled lobbies. filtering duplicates won't prevent that
I think the better solution is just invalidating them so they don't count in the first place

leia-uwu:
I just updated the test server and enabled the redis caching on it btw

leia-uwu:
bitmap

Should deleted account IDs just be excluded from the leaderboards? :thinking: 

Also i just realized we need to handle banned accounts server side too

lmssiehdev:
> I just updated the test server and enabled the redis caching on it btw  

yay

> Should deleted account IDs just be excluded from the leaderboards? 🤔

yeah.. forgot to update the delete account endpoint to remove reference to the user.. should be fixed now

leia-uwu:
81b62391d17e8614a46b5dab180a817832627f10 is a big commit so i would appreciate some reviews :p

setup.ts
```diff
@@ -0,0 +1,397 @@
...
+            initial: true,
+        });
+        if (!confirmation.value) {
+            process.exit(0);
+        }
+    }
+
+    const configText = fs.readFileSync(configPath).toString();
+    const localConfig = hjson.parse(configText);
+    util.mergeDeep(config, localConfig);
```
lmssiehdev, line 368: this fails when no config file exists (moving them inside the if statement fixed it), but besides that everything looks good, glad i won't be opening nano ever again

leia-uwu:
thoughts on changing loadout items from a json array on users table to their own table?

lmssiehdev:
> thoughts on changing loadout items from a json array on users table to their own table?

Initially they were their own separate table, moved them to the users table as optimization since the /profile route gets requested on page load and didn't want to join with the items table every time.

having them as a separate table should still be fine tho

leia-uwu:
> 50v50 mode? i have no clue how the stats looked for that.. HELP!!!

it already works as expected for match_history i think
Where theres only 2 ranks, one for each team
iirc it was like that

image
image

leia-uwu:
so there was this bug with the team menu not working randomly on localhost...

and turns out its a hono node-ws bug:
https://github.com/honojs/middleware/issues/1129

ig i will just add a setTimeout to sending the first messages on the client as a temporary workaround...
and it seems to work fine on the test server because of the latency that localhost doesn't have lol

leia-uwu:
it is done...
i'm working on a discord bot for moderation and a report system
will push them to a branch once they are somewhat usable

lmssiehdev:
ey, ngl sometimes I had doubts if this would ever get merged since it has taken SO LONG. It started as an api refactor since I was getting frustrated with how bad uws was, and it just kept growing from there.

@leia-uwu thank you for setting up the test server and making sure it's production ready 💪

PR #162, opened 2025-04-12
Polished Savannah Patches/Marksman Helmet Infinite Flares

NotASwifty13:
Resubmission of previous PR, with a touch of polishing. Also added checks to allow infinite flares if the player has the Desert Marksman Helmet equipped. 

Leia, please add the two crates I'm re-adding and the helmets still in the files to atlases (if we do patches). 

**Done:**
-Patches change by mode to be somewhat close to original, with exception of Desert (propane added).
-Cloud_01 and 02 Loot Tables Added.
-Helmet Defs Added.
-Crates and their loot tables/helmets added to MapObjectDefs. 
-Small Arms Removed from Potato Temporarily.
-Maps updated with patches, atlases, and spawn replacements.
_-Flares are infinite only when the the Desert Marksman Helmet is Equipped. Regular Endless Ammo remains the same, to remove the requirement to change GunDefs for the patch addition._ 
-SVGs for old Cloud Crates added to client > img > map.

**Reverting Back to Normal/Normal Until Release (simple, everything else can stay):**
-Remove Savannah Patches from each mode's fixed spawns. 
-Re-add Small Arms to tier_perks.


DamienVesper:
can you name your commits better please

PR #163, opened 2025-04-12
Remove extra ammo when firepower removed

### LINKED ISSUE #141: Taking off firepower perk doesnt take away extra ammo from the magazine
as title says, if you get rid of firepower perk, it will not remove your extra bullets from your weapon magazine, unlike og surviv

James46113:
Runs reload logic when firepower perk is removed from the player. This removes extra ammo that was in the gun because of the firepower perk and puts it back in the inventory.
Fixes #141

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -1052,7 +1052,12 @@ export class Player extends BaseGameObject {
         } else if (type === "fabricate") {
             this.fabricateRefillTicker = 0;
         } else if (type === "firepower") {
-            this.weaponManager.reload();
+            const originalIndex = this.weaponManager.curWeapIdx;
+            [0, 1].forEach((weapIndx) => {
+                this.weaponManager.setCurWeapIndex(weapIndx);
+                this.weaponManager.reload();
+            })
+            this.weaponManager.setCurWeapIndex(originalIndex);
         }
 
         this.recalculateScale();
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -1051,6 +1051,8 @@ export class Player extends BaseGameObject {
             }
         } else if (type === "fabricate") {
             this.fabricateRefillTicker = 0;
+        } else if (type === "firepower") {
+            this.weaponManager.reload();
         }
 
         this.recalculateScale();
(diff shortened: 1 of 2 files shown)


server/src/game/objects/player.ts
```diff
@@ -1051,6 +1051,8 @@ export class Player extends BaseGameObject {
             }
         } else if (type === "fabricate") {
             this.fabricateRefillTicker = 0;
+        } else if (type === "firepower") {
+            this.weaponManager.reload();
```
leia-uwu: This will only work for the active weapon, you need to loop over all guns instead

leia-uwu (changes requested):

server/src/game/objects/player.ts
James46113: Sorry, I'm not sure how I missed that. I'll update it

James46113: Latest commit works for both guns, I'm not sure if it's a bit hacky, but it works as intended

server/src/game/objects/player.ts
```diff
@@ -1051,6 +1051,13 @@ export class Player extends BaseGameObject {
             }
         } else if (type === "fabricate") {
             this.fabricateRefillTicker = 0;
+        } else if (type === "firepower") {
+            const originalIndex = this.weaponManager.curWeapIdx;
+            [0, 1].forEach((weapIndx) => {
+                this.weaponManager.setCurWeapIndex(weapIndx);
+                this.weaponManager.reload();
+            })
+            this.weaponManager.setCurWeapIndex(originalIndex);
```
lmssiehdev: maybe try something like this, it's less hacky.
didn't test.
```suggest
this.weaponManager.reload(GameConfig.WeaponSlot.Primary);
this.weaponManager.reload(GameConfig.WeaponSlot.Secondary);

// weaponManager.ts
reload(curWeapIdx = this.curWeapIdx): void
```

James46113:
I have left the formatting errors be as the they are not from code that I've edited, are already in master, and appear to make the code more readable.

leia-uwu:
> I have left the formatting errors be as the they are not from code that I've edited, are already in master, and appear to make the code more readable.

this is because the formatter takes line length into account, when you changed the code it changed the length of some lines so they need to be formatted again
just run `pnpm lint` locally before commiting to format it

PR #164, opened 2025-04-12
Self revive bug fix

James46113:
Fixes [bug opened on Discord](https://discord.com/channels/1278403395614408725/1359887607298789656). When the final teammate is downed, the whole team is killed even if one of the teammates has revivify. This kills the player currently reviving themselves.


James46113:
Created branch after commit, so removed previous commit that fixes an unrelated bug.

PR #167, opened 2025-04-14
Self revive bug fix

James46113:
Fixes [bug opened on Discord](https://discord.com/channels/1278403395614408725/1359887607298789656). When the final teammate is downed, the whole team is killed even if one of the teammates has revivify. This kills the player currently reviving themselves.

Resubmission of previous PR without messed up commits

### DIFF
--- a/client/src/ui/ui2.ts
+++ b/client/src/ui/ui2.ts
@@ -1617,7 +1617,7 @@ export class UiManager2 {
                 if (
                     object &&
                     player &&
-                    object == player &&
+                    (object == player || player.downed) &&
                     player.m_hasPerk("self_revive")
                 ) {
                     return this.localization.translate("game-revive-self");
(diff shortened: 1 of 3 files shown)


client/src/ui/ui2.ts
```diff
@@ -1617,7 +1617,7 @@ export class UiManager2 {
                 if (
                     object &&
                     player &&
-                    object == player &&
+                    (object == player || player.downed) &&
```
leia-uwu, line 1620: what is this fixing exactly? seems unrelated to the server fix

James46113: When a player with revivify is downed next to a downed teammate, currently the UI message says "revive teammate" but if you press "F" (or whatever keybind you use), you start reviving yourself. 

This fix ensures that when downed with revivify, the message always shows "revive self" instead of "revive teammate"

PR #168, opened 2025-04-14
Fix teammate pointer orientation

James46113:
Very minor bug fix from [issue on discord](https://discord.com/channels/1278403395614408725/1350470179041312880). Teammate pointer icon is rotated 180° to point back towards the player.

### DIFF
--- a/client/src/ui/ui.ts
+++ b/client/src/ui/ui.ts
@@ -845,7 +845,7 @@ export class UiManager {
                                 camAabb.min,
                                 camAabb.max,
                             )!;
-                            const rot = Math.atan2(dir.y, -dir.x) + Math.PI * 0.5;
+                            const rot = Math.atan2(dir.y, -dir.x) - Math.PI * 0.5;
                             const screenEdge = camera.m_pointToScreen(edge);
                             const onscreen = coldet.testCircleAabb(
                                 playerPos,
--- a/client/src/ui/ui.ts
+++ b/client/src/ui/ui.ts
@@ -845,6 +845,7 @@ export class UiManager {
                                 camAabb.min,
                                 camAabb.max,
                             )!;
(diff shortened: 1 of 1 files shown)


lmssiehdev:
Can this be fixed with css instead of modifying the client files directly?

leia-uwu:
yeah i'm also wondering what is actually broken because that logic is from the original client and it worked, so the bug must be somewhere else

James46113:
Is the asset original as well, or could that have been rotated itself?

James46113:
The styling (and therefore rotation) of the teammate indicator is set by [the code in ui.ts](https://github.com/leia-uwu/survev/blob/704a06b8aa92f01da76442443b30b3cc68541d8b/client/src/ui/ui.ts#L865-L877), so that's leading me to think it's the svg itself that is rotated compared to the original if that code is from the original client. 

Unless I'm missing rotation being set somewhere else.

lmssiehdev:
> Can this be fixed with css instead of modifying the client files directly?

thought the css files were changed up somehow, doesn't appear to be the case and it's the same as the original client

so weird

client/src/ui/ui.ts
```diff
@@ -845,7 +845,7 @@ export class UiManager {
                                 camAabb.min,
                                 camAabb.max,
                             )!;
-                            const rot = Math.atan2(dir.y, -dir.x) + Math.PI * 0.5;
+                            const rot = Math.atan2(dir.y, -dir.x) - Math.PI * 0.5;
```
leia-uwu, line 849:
```suggestion
                            // fixme: find actual cause for the indicator rotation facing backwards
                            const rot = Math.atan2(dir.y, -dir.x) - Math.PI * 0.5;
```

leia-uwu:
i added a fixme comment to note its a workaround
thanks for the PR!

PR #171, opened 2025-04-17
Added memory safe brazil

nostopgmaming17:
You're welcome


nostopgmaming17:
They dont got Segmentition Fault in Rust Saar Mam

Issue #175, opened 2025-05-01
Player can be spawned in places where they shouldn't be


wasdmco:
In Duo, second player is being spawned close to the 1st player, can lead to "problems".
![Image](https://github.com/user-attachments/assets/a2309ec1-6bfa-45a5-be8c-2a028b96fd4b)

btw: I died outside the safezone, other player didn't know how to open the vault :(

PR #178, opened 2025-05-17
Spawning Issue Fix (#175)

Edwardthe1rst:
This is a potential solution to [the spawning inside of a vault issue](https://github.com/leia-uwu/survev/issues/175), which detects when the spawning player is inside of a room with an unsolved secret and spawns him instead on the lead player. It also fixes an issue where spawning will be attempted either once or 500 (and nothing in between) times due to a missing line of code (map.ts, line ~2076: collided = false;)

This is my first time attempting to contribute to survev; any suggestions welcome.


leia-uwu:
a better fix has been pushed on commit 91b1d9f7f5d42ec61c27fe1010693676e1e13281 which should completely prevent teammates from spawning inside buildings

PR #181, opened 2025-05-21
refactor: share more code and minor clean up

lmssiehdev:


server/src/api/cache/leaderboard.ts
```diff
@@ -42,7 +44,11 @@ class LeaderBoardCache {
 
     getCacheKey(prefix: Prefix, params: LeaderboardRequest) {
         const { teamMode, mapId, type, interval } = params;
-        return `${prefix}:${teamMode}:${mapId}:${type}:${interval}`;
+        const mapInfo: MapDef = Object.values(MapDefs).find(
+            (map: MapDef) => map.mapId === mapId,
+        );
+        const mapName = mapInfo.desc.name.toLowerCase();
+        return `${prefix}:${TeamModeToString[teamMode]}:${mapName}:${type}:${interval}`;
```
leia-uwu: whats the point of this, the old logic seems fine
also could just use typescript double enum indexing to access the name: `MapId[mapId].toLowerCase()`

leia-uwu (changes requested):

server/src/api/cache/leaderboard.ts
lmssiehdev: leaderboard:1:1:most_kills:daily to leaderboard:solo:normal:...

Double indexing looks much cleaner, will fix when on laptop

PR #182, opened 2025-05-22
Add option for Local Rotation

veldreth:
This will make the game way more smoother, because of the client-side mouse move if its enabled from settings (default: true). Using also the same old logic if disabled which is synced and looks bad in some cases.


client/src/config.ts
```diff
@@ -43,6 +43,7 @@ const defaultConfig = {
     highResTex: true,
     interpolation: true,
     screenShake: true,
+    LocalRotation: true,
```
leia-uwu, line 46: first letter should be lowercase

client/src/objects/player.ts
```diff
@@ -188,6 +190,8 @@ export class Player implements AbstractObject {
     __id!: number;
     __type!: ObjectType.Player;
     active!: boolean;
+    isSpectating: boolean = false;
+    activeId: number = -1;
```
leia-uwu, line 194: these variables are never set anywhere... have you even tested this code? because from reading it it shouldn't even work at all
and you shouldn't be storing those 2 values anyway, just pass it down the function

client/src/objects/player.ts
```diff
@@ -56,6 +57,7 @@ import type { Obstacle } from "./obstacle";
 import type { Emitter, ParticleBarn } from "./particles";
 import { halloweenSpriteMap } from "./projectile";
 import { createCasingParticle } from "./shot";
+const inputManager = new InputHandler(document.body);
```
leia-uwu, line 60: this is just hard no.
don't initialize a new global instance just for this, use the one the game instance already has

leia-uwu (changes requested):
Code comments aside, i dont like local rotation being a boolean and on by default

it looks completely terrible with high ping, example in suroi:
https://github.com/user-attachments/assets/2f6ac92f-4403-45a1-bee7-7bd74a436c96

what i would like is it being automatic based on the ping, so it gets disabled if the ping is above a threshold to make the gap between the server and client lower

client/src/objects/player.ts
veldreth: i already tested the code this with activeid is used for the check down + mobile if its not initialized it wont work because on phone is not ok, and the game like now its not really smooth I tested it a lot of times, I saw almost the same method in other private servers as zurvev, resurviv, the logic is that and is toggleable. People can disable/enable.

veldreth: I did it for couple of minutes thats why I messed up with some parts, currently tested out - all works fine and without problems/errors. This variables if are missing:
    isSpectating: boolean = false;
    activeId: number = -1;
    it won't work I need them for here at Zr {:
    if (this.activeId == this.__id && !this.isSpectating && device.mobile == false && camera.m_localrotationEnabled) {

veldreth: Even if you make it with ping check if high just not to work and not be local it will be better, that was just my idea, of course it can be way more better.

DamienVesper: > i already tested the code this with activeid is used for the check down + mobile if its not initialized it wont work because on phone is not ok, and the game like now its not really smooth I tested it a lot of times, I saw almost the same method in other private servers as zurvev, resurviv, the logic is that and is toggleable. People can disable/enable.

This does not address the original issue. Your proposal unnecessarily creates an additional `InputHandler` instead of reusing the provided one.

martink1337: As i said I did it for couple of minutes in a rush just to diagnose if from there is the laggy movement, that part can be removed and just  the inputhandler without second const - OF COURSE, its subject to change, I just gived the idea, because it will be way more better.

DamienVesper: If your PR is not ready, please mark as draft. Otherwise, do not make PR. Code changes are requested by Leia above. Please fix and resubmit.

veldreth: I'll make an resolve as I said I did it very fast and didn't saw in that moment the problem

veldreth: instead of creating this inputmanager we can add in player class
class Player {
  game: Game;

  constructor(game: Game) {
    this.game = game;
  }
}

and constructor game: Game
this.game = game
and do const mouseX = this.game.m_input.mousePos.x;
const mouseY = this.game.m_input.mousePos.y;

(already done in game )

or easier way if it will work (here im not completely sure)
const mouseX = this.scene.game.m_input.mousePos.x;
const mouseY = this.scene.game.m_input.mousePos.y;

- for LocalRotation, I started writing it with upper, this can be just lowered to anywhere , does't matter.
Only this:
    isSpectating: boolean = false;
    activeId: number = -1;
    should stay because of the check down in Zr, if they are removed as you say, it won't work, test it out and you'll find out if you think it is a problem.

leia-uwu (commented):
?????????????????????????????????????????????????????????????

leia-uwu:
yeah not merging this
the code does not work, is bad, and now theres this random ass commit renaming everything to TESTsurviv.io??
what.

lmssiehdev:
> testsurviv.io

made me lol

PR #193, opened 2025-06-10
50v50 Updates 

NotASwifty13:
Finally getting this in..... Feel free to cherry pick if necessary. 

**New Role - Captain:**
-Promoted on leader's death.
-Also given leader ping ability.
-Granted Assume Leadership, 8x Scope, Lvl. 4 Captain Helmet, and half adrenaline on promotion.
(Lone Survivr Promotion should occur if a player is eligible for both roles).
(Credit to Palyer for sound!).

**New Perk - Assume Leadership:**
-Slight increase in size, smaller than Leadership.
-Constantly half full adrenaline, instead of full like Leadership.

**Preload Guns Changes:**
Can now preload guns using one of three methods:
-Add "preload: true" to a specific gun in a tier (e.g. AWM-S in Golden Drops).
-Preload an entire crate (the old J&N method) -- Should use for autoLoot items as well (like DEagle Case).
-Preload an entire mode's loot (set to 50v50 right now, but others could be added).

**Auto Fire Flare Gun:**
-Leaders receive the flare/gun as normal. 
-Given 15 seconds to fire it. If not it auto fires, and the flare ammo is removed from the leader, whether it is loaded in the flare gun or not. 
-Leaders cannot drop flare ammo. They can drop the flare gun.

**Fourth Perk UI:**
-Self Explanatory 

**Iron Bomb Sprite:**
-Fixed up version of the new sprite I made and sent a few months ago.

AP Rounds is being worked on, but needs more discussion.


server/src/game/objects/bullet.ts
```diff
@@ -124,6 +124,15 @@ export class BulletBarn {
...
 
+        if (params.bulletType === "bullet_flare" && params.playerId) {
+            const player = this.game.playerBarn.players.find(
+                (p) => p.playerId === params.playerId,
+            );
+            if (player && player.role === "leader") {
+                player.hasFiredFlare = true;
+            }
+        }
+
```
leia-uwu, line 135: imo bullet.ts is not the right place for this

and looping over all players to find one by ID is unnecessarily slow,  `game.objectRegister.getById` is faster since it doesn't require a loop

server/src/game/objects/player.ts
```diff
@@ -298,6 +298,14 @@ export class PlayerBarn {
                             util.randomInt(0, promotablePlayers.length - 1)
                         ];
                     randomPlayer.promoteToRole(scheduledRole.role);
+
+                    if (scheduledRole.role === "leader" && randomPlayer.team) {
+                        setTimeout(() => {
+                            if (randomPlayer.team && !randomPlayer.hasFiredFlare) {
+                                randomPlayer.team.autoCallAirdrops();
+                            }
+                        }, 15000);
```
leia-uwu, line 307: dont use setTimeouts

server/src/game/objects/player.ts
```diff
@@ -1484,6 +1498,11 @@ export class Player extends BaseGameObject {
                         target.setDirty();
                         target.setGroupStatuses();
                         this.game.pluginManager.emit("playerRevived", target);
+
+                        if (target.hasPerk("assume_leadership")) target.boost = 50;
+                        target.setDirty();
+                        target.setGroupStatuses();
+                        this.game.pluginManager.emit("playerRevived", target);
```
leia-uwu, line 1505: you duplicated the `setDirty`, `setGroupStatuses` and emitting the event for no reason

server/src/game/team.ts
```diff
@@ -70,6 +72,35 @@ export class Team {
...
+        if (flareGun) {
+            flareGun.ammo = 0;
+            this.leader.weapsDirty = true;
+        }
+
+        if (this.leader.inventory[ammoType]) {
+            this.leader.inventory[ammoType] = 0;
+            this.leader.inventoryDirty = true;
+        }
+    }
```
leia-uwu, line 103: `gameSourceType` being `"game"` is just ??? :sob:, where did that came from, it should be a valid item ID, and in this case, `"flare_gun"` lol

and setting the flare ammo to 0 on the inventory is also wrong, if the leader has 2 flare ammo it will remove both while only firing one

also what if they got dual flare gun instead of single, then it wont set the ammo to 0 on the gun in the inventory, creating effectively a single time per game airdrop dupe exploit

PR #195, opened 2025-06-11
50v50 Update (Revised) 

NotASwifty13:
Fixes done, except removing the check for manual flare gun fire from bullet.ts. PR's a bit more messy now unfortunately. 

New Role - Captain:
-Promoted on leader's death.
-Also given leader ping ability.
-Granted Assume Leadership, Leader Outfits, 8x Scope, Lvl. 4 Captain Helmet, and half adrenaline on promotion.
(Lone Survivr Promotion should occur if a player is eligible for both roles).
(Credit to Palyer for sound!).

New Perk - Assume Leadership:
-Slight increase in size, smaller than Leadership.
-Constantly half full adrenaline, instead of full like Leadership.

Preload Guns Changes:
Can now preload guns using one of three methods:
-Add "preload: true" to a specific gun in a tier (e.g. AWM-S in Golden Drops).
-Preload an entire crate (the old J&N method) -- Should use for autoLoot items as well (like DEagle Case).
-Preload an entire mode's loot (set to 50v50 right now, but others could be added).

Auto Fire Flare Gun:
-Leaders receive the flare gun as usual. They cannot drop it.
-If they fire the gun themselves, they can drop the gun. 
-If they don't, the flare will automatically fire 15s after promotion (even in a building). They can then drop the flare gun.
-If the flare gun is picked up again, it can be dropped without requiring it to be fired. 

Fourth Perk UI:
-Self Explanatory

Iron Bomb Sprite:
-Removed.

AP Rounds:
-Additional 1.5 multiplier applied to obstacles. 
-Damage multiplier that increases according to opponent's vest. Starts at 1.15 for "", and ends at 1.35 for "chest_04".
-Replaces splinter rounds on Lone Survivrs.
(Thank you headphones pig for the sprite!).


server/src/game/objects/bullet.ts
```diff
@@ -594,6 +606,21 @@ export class Bullet {
                     if (isHighValueTarget) {
                         multiplier *= 1.25;
                     }
+                    // AP Rounds Vest Multipliers. Subject to balance changes.
```
NotASwifty13: Fixed AP Rounds (Finally)...

PR #197, opened 2025-06-12
Server fixes

cytronicoder:
Added a util function `waitForApiServerReady` to check that the game server does not attempt to communicate with the API server before it is fully available. This change should prevent issues where the game server would fail or behave unexpectedly due to early requests to an unready API.

Aforementioned function was added to the game server startup sequence—just before it begins sending data. All game server processes relying on API availability are now only initiated when the API is confirmed to be responsive.

Also, `dev:api` and `dev:game` scripts in `server/package.json` were updated to ignore changes in `node_modules` to reduce unnecessary rebuilds during development.


cytronicoder:
Hi, thanks for your reply. Your feedback makes sense; I just had some strange behavior where the server would randomly restart even when no changes had been made to the code, but fair point on the dev flow impediment.

As to your comments about vibe coding, I wasn’t intending to come off like I was just throwing stuff at the wall, so cheers! Have a good day

PR #198, opened 2025-06-15
Update stats.ts

Dread-panda:
The schema would return NaN for mapid if it was not a number, which resulted in an error when trying to retrieve values from the database (mapid had to be int). All this did was return an error 500 instead of 400, but this should fix it. Also first time doing a github fork/push thing so sorry if i did this wrong

### DIFF
--- a/shared/types/stats.ts
+++ b/shared/types/stats.ts
@@ -130,7 +130,8 @@ export const zLeaderboardsRequest = z.object({
     mapId: z
         .string()
         .min(1)
-        .transform((v) => Number(v)),
+        .transform((v) => Number(v))
+        .transform((n) => isNaN(n) ? 0 : n),
     type: z.enum(["most_kills", "most_damage_dealt", "kpg", "kills", "wins"]),
     // why tf is this sent as a string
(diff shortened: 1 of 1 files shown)


lmssiehdev:
> The schema would return NaN for mapid if it was not a number

this is the intended behavior to match the client, since it send an actual string instead of an int
if the string isn't a valid number, zod should return with 400 

the mapId also needs to be validated
```ts
mapId: z.enum(VALID_MAP_IDS).transform((v) => Number(v)),
```

Dread-panda:
It does not return 400 though, it returns 500 because if it is not valid it returns NaN, which eventually gets inserted into the sql database (which is why it returns 500 instead of 400). Try it.

Dread-panda:
Might sound silly, but don't the rest of the parameters in that function gets defaulted to something with the .catch()? Timeframe defaults to daily and mode defaults to solo, so should mapID not default to 0 to make it consistent?

lmssiehdev:
I made a pr to fix that at #200, thanks for the report and for testing the validation.

PR #202, opened 2025-06-24
fix: lint ci & fix biome version

lmssiehdev:
https://biomejs.dev/internals/versioning/#why-pinning-the-version-is-important


.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Setup Biome
         uses: biomejs/setup-biome@v2
         with:
-          version: latest
+          version: "2.0.5"
```
leia-uwu: from https://github.com/marketplace/actions/setup-biome :

>    The version of the Biome CLI to install.
>    This input is optional and by default the version will be automatically
>    detected from the project's dependencies. If no version is found in the
>    project's dependencies, the latest version of the Biome CLI will be installed.
>    Example values: "2.0.4", "latest"

so ig we can just remove the version from the CI so it uses the one from the package json to avoid it being duplicated?

PR #205, opened 2025-06-29
refactor: descriptive error messages

lmssiehdev:
didn't add any description to the localization files so the client falls back to showing the default error msg, this is intentional.


server/src/api/index.ts
```diff
@@ -1,5 +1,6 @@
 import { serve } from "@hono/node-server";
 import { createNodeWebSocket } from "@hono/node-ws";
+import { error } from "console";
```
leia-uwu: accidental autocompletion import? lol

PR #207, opened 2025-06-30
feat: minimal moderation bot

lmssiehdev:
minimal as in wanting the api to do all the work when possible, most commands would be just an rpc call to the endpoint and returning it's message.

no premature optimization was done, until i get enough feedback on how this would be used, so some duplication and hardcoded things are there.

### DIFF
--- a/bot/package.json
+++ b/bot/package.json
@@ -7,10 +7,8 @@
     "dev": "tsx watch --clear-screen=false --include '../server/src/api/routes/private/*' --include ../survev-config.hjson src/index.ts",
     "register-commands": "tsx src/register-commands.ts"
   },
-  "devDependencies": {
-    "tsx": "^4.19.4"
-  },
   "dependencies": {
+    "tsx": "^4.19.4",
     "discord.js": "^14.21.0",
     "hono": "^4.7.11",
     "zod": "^3.25.45"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -42,13 +42,12 @@ importers:
       hono:
         specifier: ^4.7.11
         version: 4.7.11
-      zod:
-        specifier: ^3.25.45
-        version: 3.25.45
-    devDependencies:
       tsx:
         specifier: ^4.19.4
         version: 4.19.4
+      zod:
+        specifier: ^3.25.45
+        version: 3.25.45
 
   client:
     dependencies:


bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,85 @@
...
+import type { InferResponseType } from "hono";
+import { createDiscordDropdownUI } from "../components";
+import { Command, honoClient } from "../utils";
+
+export type SelectedPlayer = Extract<
+    InferResponseType<typeof honoClient.moderation.get_player_ip.$post, 200>,
+    any[]
+>[number];
+
+declare module "discord.js" {
```
DamienVesper: what the fuck

lmssiehdev: fuck fuck fuckity fuck

server/src/api/db/drizzle/0002_brown_lizard.sql
leia-uwu, line 1: maybe squash this and team_mode schema changes into the same file?

bot/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "dev": "tsx watch --clear-screen=false --include ../server/src/api/routes/private/private.ts  src/index.ts",
+    "register-commands": "tsx src/register-commands.ts"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
leia-uwu: GPL instead?

bot/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "packageManager": "pnpm@10.11.0",
```
leia-uwu: pretty sure this should only be on the root package json

bot/src/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+
+export const honoClient = hc<PrivateRouteApp>(API_URL, {
+    headers: {
+        "survev-api-key": Config.secrets.SURVEV_API_KEY,
+    },
+});
+
+export function hasPermission(interaction: Interaction): boolean {
+    if (interaction.memberPermissions?.has(PermissionFlagsBits.Administrator)) {
```
leia-uwu, line 37: shouldn't we also check if the guild ID is correct as a sanity check?
from what i understood it only register the commands on the specific guild so people cant just add the bot to their own server and use it but i'm too paranoid to trust just that lol

setup.ts
```diff
@@ -349,6 +360,23 @@ async function setupDevelopmentConfig(config: PartialConfig) {
...
+        type: "text",
+    });
+    config.secrets.DISCORD_BOT_TOKEN = discordBotToken.value;
+
+    const discordRoleId = await prompt<{ value: string }>({
+        message: "Enter the discord role ID",
+        name: "value",
+        type: "text",
+    });
+    config.discordRoleId = discordRoleId.value;
```
leia-uwu: Missing guild id

bot/src/commands/helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+    const res = await honoClient.moderation[name].$post({
+        json: args.data as any,
+    });
+
+    const { message } = await res.json();
+
+    await interaction.reply(message);
+}
+
+export function createSlashCommnad(config: ReturnType<typeof createCommand>) {
```
DamienVesper: (Lint): spellcheck (and all invocations of this function across the PR).

bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,66 @@
...
+            option
+                .setName("in-game-name")
+                .setDescription("The name of the player to search for")
+                .setRequired(true),
+        ),
+
+    async execute(interaction: ChatInputCommandInteraction) {
+        const searchName = interaction.options.getString("in-game-name")!;
+
+        await interaction.deferReply({});
```
DamienVesper: `interaction.deferReply();` does the same thing. The argument is optional.

bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        try {
+            const res = await honoClient.moderation.get_player_ip.$post({
+                json: {
+                    name: searchName!,
+                },
+            });
+
+            if (!res.ok) {
+                await interaction.editReply({
```
DamienVesper, line 40: This and any subsequent `interaction#editReply` should be `interaction#followUp` instead.
You haven't actually sent a reply with `interaction#reply` yet. You have only sent an initial deferred response, which is not editable.

bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,66 @@
...
+                    content: "Failed to get player IP",
+                });
+                return;
+            }
+
+            const result = await res.json();
+
+            // no player found, return message formatted by the api;
+            if ("message" in result) {
+                await interaction.editReply({
```
DamienVesper, line 50: `interaction#editReply` -> `interaction#followUp`

bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,66 @@
...
+                });
+                return;
+            }
+
+            const matchingPlayers = result;
+
+            await createDiscordDropdownUI(interaction, matchingPlayers, searchName);
+        } catch (error) {
+            console.error("Error in banplayer command:", error);
+            await interaction.editReply({
```
DamienVesper, line 61: `interaction#editReply` -> `interaction#followUp`

bot/src/commands/search-player.ts
```diff
@@ -0,0 +1,66 @@
...
+
+            // no player found, return message formatted by the api;
+            if ("message" in result) {
+                await interaction.editReply({
+                    content: result.message,
+                });
+                return;
+            }
+
+            const matchingPlayers = result;
```
DamienVesper, line 56: Unnecessary assignment of existing object to another variable. Just pass result in `createDiscordDropdownUI`.

bot/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { type Command, hasPermission } from "./utils";
+
+const client = new Client({ intents: [GatewayIntentBits.Guilds] });
+
+function setupInteractionHandlers() {
+    client.on(Events.InteractionCreate, async (interaction) => {
+        if (!hasPermission(interaction)) {
+            if (interaction.isRepliable()) {
+                await interaction.reply({
+                    content: "You do not have permission to use this command",
```
DamienVesper: Keep in mind this executes for all `Exclude<Interaction<Cached>, AutocompleteInteraction<Cached>>`. These interactions do not necessarily have to be commands.

bot/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+        const commandName = interaction.commandName as Command;
+        if (!commandHandlers[commandName]) {
+            console.warn(`Unknown command: ${commandName}`);
+            return;
+        }
+        try {
+            await commandHandlers[commandName](interaction);
+        } catch (error) {
+            console.error(`Error executing command "${commandName}":`, error);
+            await interaction.reply({
```
DamienVesper: Obviously, this is dependent on command implementation, but it would be preferable to check whether the interaction has been deferred and then using either `interaction#reply` or `interaction#followUp` with the appropriate flags.

bot/src/config.ts
```diff
@@ -0,0 +1,30 @@
+import { getConfig } from "../../config";
+
+const isProd = process.env["NODE_ENV"] === "production";
```
DamienVesper, line 3: (style): Please use `process.env.NODE_ENV` instead.

bot/src/config.ts
```diff
@@ -0,0 +1,30 @@
...
+if (
+    !Config.discordGuildId ||
+    !Config.discordRoleId ||
+    !Config.secrets.DISCORD_CLIENT_ID ||
+    !Config.secrets.DISCORD_BOT_TOKEN
+) {
+    throw new Error("Bot config not set up properly");
+}
+
+const API_URL = `http://${Config.apiServer.host}:${Config.apiServer.port}/private`;
```
DamienVesper: The protocol can be either `http` or `https`. You should just use `Config.gameServer.apiServerUrl` instead.

bot/src/commands/helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+    const builder = new SlashCommandBuilder()
+        .setName(config.name)
+        .setDescription(config.description);
+
+    for (const option of config.options) {
+        switch (option.type) {
+            case ApplicationCommandOptionType.String:
+                builder.addStringOption((stringOption) =>
+                    stringOption
+                        .setName(option.name as string)
```
DamienVesper, line 69: Willful wishing. Maybe try discriminated unions?

bot/src/register-commands.ts
```diff
@@ -0,0 +1,25 @@
+import { REST, Routes } from "discord.js";
+import { commandsToRegister } from "./commands";
+import { DISCORD_BOT_TOKEN, DISCORD_CLIENT_ID, DISCORD_GUILD_ID } from "./config";
+
+async function run() {
+    async function registerCommands() {
+        const rest = new REST({ version: "10" }).setToken(DISCORD_BOT_TOKEN);
+        try {
+            console.log("Started refreshing application (/) commands.");
```
DamienVesper: Maybe change these from the discordjs.guide defaults.

bot/src/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+export function hasPermission(interaction: Interaction): boolean {
+    if (interaction.guild?.id !== DISCORD_GUILD_ID) {
+        return false;
+    }
+
+    if (interaction.memberPermissions?.has(PermissionFlagsBits.Administrator)) {
+        return true;
+    }
+
+    const member = interaction.member as GuildMember;
```
DamienVesper: You should the use `interaction#inCachedGuild()` type guard instead of overriding types.

bot/src/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    if (member && "roles" in member) {
```
DamienVesper: This check is unnecessary if you use the aforementioned type guard.

DamienVesper:
> command names should be the same as API endpoint
> 
> we use endpoint params to declare the command options, and since discord doesn't accept camelCase names, going forward they should be in snake_case

In my opinion these choices make the slash commands comestically ugly (no underscores is preferable). But this is not really an issue for the usability of the bot as a whole, or the people who will be making use of it anyway.

bot/src/commands/helpers.ts
lmssiehdev: we don't care about the type here, narrowing this would just unnecessarily overcomplicate the types

bot/src/config.ts
lmssiehdev: You're right, but I copied this from elsewhere in the codebase, so this is our style :p

Not gonna fix it now, but we should and  this is the linter's job, and I remember biome used to fix before.

bot/src/register-commands.ts
lmssiehdev: I do not see the issue with them being from docs, feel free to edit them to your liking tho.

bot/src/commands/search-player.ts
lmssiehdev: it's descriptive, gonna keep it until `createDiscordDropdownUI` starts accepting an object instead

DamienVesper (approved):

bot/src/config.ts
```diff
@@ -0,0 +1,34 @@
...
+    !Config.discordRoleId ||
+    !Config.secrets.DISCORD_CLIENT_ID ||
+    !Config.secrets.DISCORD_BOT_TOKEN ||
+    !Config.gameServer.apiServerUrl
+) {
+    throw new Error("Bot config not set up properly");
+}
+
+const API_URL = isProd
+    ? Config.gameServer.apiServerUrl
```
leia-uwu: uh missing /private here...
also what even is the point of this ternary, cant it just always be `${Config.gameServer.apiServerUrl}/private` regardless of prod / dev?

server/src/api/routes/private/ModerationRouter.ts
```diff
@@ -11,70 +12,106 @@ import { db } from "../../db";
...
+export const zSetMatchDataNameParams = z.object({
+    current_name: z.string(),
+    new_name: z.string(),
+});
+
+export const zSetAccountNameParams = z.object({
+    new_name: z.string(),
+    current_slug: z.string(),
+});
+
```
leia-uwu: maybe those should be moved to their own file because importing whole `ModerationRouter.ts` on the bot causes it to import way more stuff and also load the config file twice since it will be importing the server `config.ts` too

PR #209, opened 2025-07-05
feat: button to confirm all items in loadout

lmssiehdev:
the client hijacks the model hide fn to confirm and display the next time, so couldn't find an elegant way to implement the confirm all, ig this will do.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -1429,6 +1429,7 @@
             </div>
             <div class="modal-footer modal-footer-name modal-footer-round">
               <h3 class="close close-footer" data-l10n="index-confirm">Confirm</h3>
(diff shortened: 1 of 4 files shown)


client/index.html
```diff
@@ -1429,6 +1429,7 @@ <h2 id="modal-account-name-title">You got a new item!</h2>
             </div>
             <div class="modal-footer modal-footer-name modal-footer-round">
               <h3 class="close close-footer" data-l10n="index-confirm">Confirm</h3>
+                <h3 class='close close-footer' data-l10n='index-confirm-all' data-confirm-all="true">Confirm All</h3>
```
leia-uwu: wrong indentation
tbh we need to set up a formatter for the html files

PR #214, opened 2025-07-22
Fixes

NotASwifty13:


DamienVesper (changes requested):
- avoid idstring checking if possible
 - duplicated logic. please extract to common method.

PR #222, opened 2025-07-25
refactor (net): clean up network code a bit

leia-uwu:
nothing seems to have broken so far, tested a full 50v50 match with stress test bots and no packet errors...


DamienVesper:
the prototype method spam for perf is crazy

PR #228, opened 2025-07-29
fix: fix ip_bans command hashing ips twice

lmssiehdev:


shared/types/moderation.ts
```diff
@@ -53,3 +53,9 @@ export const zSetAccountNameParams = z.object({
 export const zFindDiscordUserSlugParams = z.object({
     discord_user: z.string(),
 });
+
+export const zGiveItemParams = z.object({
+    item: z.string(),
+    slug: z.string(),
+    source: z.string().default("daddy-has-privileges"),
```
leia-uwu: lol i think we should just change to unlock_default atp :p

lmssiehdev: :(

lmssiehdev: i'm opposed to this change, do it yourself :)

leia-uwu (approved):

PR #230, opened 2025-08-04
enhance: remove confusing characters from team codes

1092384:


1092384:
The "confusing characters" in question are capital o, the number 0, uppercase i, and lowercase L.

1092384:
This is useful for players that are playing with each other IRL where it might be faster to enter in 4 characters than sending a link

server/src/teamMenu.ts
```diff
@@ -348,13 +348,11 @@ class Room {
     }
 }
 
-const alphanumerics = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+const teamCodeCharacters = "ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz123456789";
 function randomString(len: number) {
```
DamienVesper: If you are doing this should extract this to a seperate custom function that is not `randomString`, given that you are making this change __specifically__ for team codes.

lmssiehdev: function isn't exported, and i don't think it's used for anything besides the room url

DamienVesper: Well I guess that's a potential refactor in itself, isn't it.
I guess can keep then? Still don't think this is a real solution as changing the font would be better option (as per @IOMAN1 ).

lmssiehdev:
@1092384 please run `pnpm run lint` before pushing

lmssiehdev:
LGTM!

PR #233, opened 2025-08-04
Airstrikes

IOMAN1:


server/src/game/objects/plane.ts
```diff
@@ -52,9 +49,8 @@ export class PlaneBarn {
     }[] = [];
     airstrikeZones: AirstrikeZone[] = [];
 
-    specialAirdrop = {
-        canDrop: false,
-        dropped: false,
+    helpForLosingTeam = {
+        sent: false,
```
leia-uwu: doesn't need to be an object now that it has only 1 key

leia-uwu (dismissed):
minor nitpick but LGTM

PR #241, opened 2025-08-07
minor QOL tweaks

IOMAN1:


server/src/game/objects/player.ts
```diff
@@ -891,23 +893,37 @@ export class Player extends BaseGameObject {
...
                 this.setOutfit(newOutfit(clampedTeamId));
             } else {
                 // string
-                if (newOutfit) this.setOutfit(newOutfit);
+                if (newOutfit) {
+                    if (!oldOutfit.noDropOnDeath) {
+                        this.dropLoot(this.outfit);
+                    }
+                    this.setOutfit(newOutfit);
+                }
```
leia-uwu: nitpick: should just extract newOutfit to a variable and set / drop the old one outside the if statement to avoid duplicating that

PR #246, opened 2025-08-12
Debug menu rewrite

lmssiehdev:
now using [tweakpane](https://tweakpane.github.io/docs/)
and added a debug HUD with graphs and stuff

### DIFF
--- a/client/src/game.ts
+++ b/client/src/game.ts
@@ -12,7 +12,11 @@ import { Camera } from "./camera";
 import type { ConfigManager, DebugRenderOpts } from "./config";
 import { DebugHUD } from "./debug/debugHUD";
 import { debugLines } from "./debug/debugLines";
+
+/* STRIP_FROM_PROD_CLIENT:START */
 import { Editor } from "./debug/editor";
+/* STRIP_FROM_PROD_CLIENT:END */
+
 import { device } from "./device";
 import { EmoteBarn } from "./emote";
 import { Gas } from "./gas";


client/src/game.ts
```diff
@@ -9,10 +9,15 @@ import { v2 } from "../../shared/utils/v2";
 import type { Ambiance } from "./ambiance";
 import type { AudioManager } from "./audioManager";
 import { Camera } from "./camera";
-import type { ConfigManager, DebugOptions } from "./config";
-import { debugLines } from "./debugLines";
+import type { ConfigManager, DebugRenderOpts } from "./config";
+import { DebugHUD } from "./debug/debugHUD";
+import { debugLines } from "./debug/debugLines";
+
+/* STRIP_FROM_PROD_CLIENT:START */
```
DamienVesper, line 16: As of right, now the injection method is very messy, stripping imports in production and using several `IS_DEV` checks. Should later be refactored to 1 check and lazy-loading with dynamic import, which is the 'preferred' method to prevent stuff from being included in production bundle.

client/src/debug/debugHUD.ts
```diff
@@ -0,0 +1,234 @@
...
+        this.graphsContainer.addChild(graph.container);
+        this.graphs.push(graph);
+        return graph;
+    }
+
+    addLine(text: string) {
+        let line = this.infoLines.find((l) => !l.visible);
+        if (!line) {
+            line = new PIXI.Text();
+            line.style = { ...defaultLabelTextOptions };
```
DamienVesper: The style is equivalent to `defaultLabelTextOptions`, so maybe you can do pass-by-reference instead of pass-by-value here?

server/src/game/objects/player.ts
```diff
@@ -1259,15 +1257,18 @@ export class Player extends BaseGameObject {
...
+        speedEnabled: false,
+        speed: 1,
+
+        noClip: false,
+        godMode: false,
+
         /** drag and drop loot, obstacles, and buildings */
-        godMode: <GodMode>{
-            isGodMode: false,
+        moveObjMode: <MoveObjsMode>{
```
DamienVesper, line 1270: Terrible name, should use [this](https://gmod.fandom.com/wiki/Physgun) instead.

client/src/debug/debugHUD.ts
```diff
@@ -0,0 +1,234 @@
...
+        const cfg = this.hudConfig();
+
+        if (cfg.fps.show || cfg.fps.showGraph) {
+            this.frames.push(dt);
+            this.fpsTicker += dt;
+            // add an average FPS every 100 milliseconds
+            // adding an entry every frame makes the graph go crazy and become kinda useless
+            // specially on high refresh rates
+            if (this.fpsTicker > 0.1) {
+                const sum = this.frames.reduce((a, b) => a + b, 0);
```
DamienVesper, line 123: this is slower than a reverse for loop but that's a nitpick

DamienVesper (dismissed):
Once again, I have no idea what code I'm reading.

server/src/game/objects/player.ts
leia-uwu: erm no

client/src/game.ts
leia-uwu: as discussed on discord, using async and dynamic imports on the client currently is a PITA because nothing is async there so will leave that for the future

PR #251, opened 2025-08-19
 fix: serving static files from the wrong route

lmssiehdev:
results in a 404 when navigating to the main site

### DIFF
--- a/HOSTING.md
+++ b/HOSTING.md
@@ -75,7 +75,7 @@ npm i -g pnpm
 IF you want to have accounts, leaderboards and IP bans, you will have to install and set-up PostgreSQL database
 
 ```sh
-sudo apt -y postgresql
+sudo apt install -y postgresql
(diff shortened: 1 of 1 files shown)


HOSTING.md
```diff
@@ -75,8 +75,7 @@ npm i -g pnpm
 IF you want to have accounts, leaderboards and IP bans, you will have to install and set-up PostgreSQL database
 
 ```sh
-sudo apt -y postgresql
-sudo -u postgres initdb --locale=C.UTF-8 --encoding=UTF8 -D /var/lib/postgres/data --data-checksums
+sudo apt install -y postgresql
```
DamienVesper, line 78: Never a fan of putting `-y` in docs. People should always read the proposed changes to their system if they don't know what they're doing.

DamienVesper (approved):
other than that LGTM

PR #259, opened 2025-08-28
refactor: add inventoryManager class

leia-uwu:
This cleans up server inventory code a lot and removes a lot of duplicated code

And also makes it more safe and less bug prone


leia-uwu:
~~eipi reviewing it on VC trust its good~~

PR #262, opened 2025-09-06
Feat/dynamic client theme

leia-uwu:


client/vite.config.mts
```diff
@@ -8,44 +8,11 @@ import { atlasBuilderPlugin } from "./atlas-builder/vitePlugin";
...
-    },
-    easter: {
-        menuMusic: "audio/ambient/menu_music_01.mp3",
-        splashBg: "/img/main_splash_easter.png",
-    },
-    halloween: {
-        menuMusic: "audio/ambient/menu_music_02.mp3",
-        splashBg: "/img/main_splash_halloween.png",
-    },
-    faction: {
```
lmssiehdev, line 24: shouldn't the custom menu be assigned elsewhere since this is removed?

leia-uwu: making that dynamic seemed 20 times harder than the BG and its only used for halloween and i didnt want hex to murder me further

lmssiehdev: fair

PR #264, opened 2025-09-16
feat (server/client): added featured youtubers and twitch streamers 

NAMERIO:
- Updated ApiServer to fetch and cache twitch streamer data via Helix API
- Twitch OAuth tokens expire every 60 days and must be refreshed, but we can probably make it refresh itself automatically in the future
- also added Youtube rotation support so multiple youtubers can display


server/package.json
```diff
@@ -23,13 +23,15 @@
...
     "@oslojs/crypto": "^1.0.1",
+    "@types/node-fetch": "^2.6.13",
     "arctic": "^3.7.0",
     "better-sqlite3": "^11.10.0",
     "croner": "^9.1.0",
     "dotenv": "^16.6.1",
     "drizzle-orm": "^0.43.1",
     "hono": "^4.9.7",
     "nanotimer": "^0.3.15",
+    "node-fetch": "^2.7.0",
```
leia-uwu: ???????
are we in 2018

nodejs has had fetch support for years and we already had fetch usage everywhere in the server without a package

why the hell did you add that

client/src/siteInfo.ts
```diff
@@ -142,13 +148,15 @@ export class SiteInfo {
             );
 
             const featuredYoutuberElem = $("#featured-youtuber");
-            const displayYoutuber = this.info.youtube;
-            if (displayYoutuber) {
+            if (this.info.youtube && this.info.youtube.length > 0) {
+                const yt = this.info.youtube[Math.floor(Math.random() * this.info.youtube.length)];
```
leia-uwu: just keep the request as a single item like it was and make the randomization server side...

client/src/siteInfo.ts
```diff
@@ -34,11 +34,17 @@ export class SiteInfo {
...
+        const fetchSiteInfo = () => {
+            $.ajax(siteInfoUrl).done((data: SiteInfoRes) => {
+                this.info = data || {};
+                this.loaded = true;
+                this.updatePageFromInfo();
+            });
+        };
+
+        fetchSiteInfo();
+        setInterval(fetchSiteInfo, 300000);
```
leia-uwu: i dont really like that....
i was thinking about refreshing the site info once the player leaves the game to go to the main menu if it has passed enough time

because that timer can trigger refreshes mid game which could cause lag spikes in bad connections

configType.ts
```diff
@@ -404,6 +421,20 @@ export interface ConfigType {
          */
         inventory?: Record<string, number>;
     };
+
+    /**
+     * Featured streamers and youtubers to show on the main menu.
+     * These rotate randomly via the API.
```
leia-uwu, line 427: me when that comment is just a lie because you made it client side for youtubers

server/src/api/apiServer.ts
```diff
@@ -111,6 +115,48 @@ export class ApiServer {
...
+    }
+
+    private async fetchTwitchData(name: string) {
+        try {
+            const res = await fetch(
+                `https://api.twitch.tv/helix/streams?user_login=${name}`,
+                {
+                    headers: {
+                        "Client-ID": Config.secrets.TWITCH_CLIENT_ID ?? "",
+                        Authorization: `Bearer ${Config.secrets.TWITCH_OAUTH ?? ""}`,
```
leia-uwu: this method should just early return null if we have no twitch server keys, removing the need for `?? ""`
no reason to make a request that will always fail from missing keys

config.ts
```diff
@@ -34,6 +34,21 @@ export function getConfig(isProduction: boolean, dir: string) {
...
+                { name: "RazerTiduu", link: "https://www.youtube.com/@razertiduu1876" },
+                { name: "squiggle", link: "https://www.youtube.com/@squiggle" },
+                { name: "this.is.gls_", link: "https://www.youtube.com/@this.is.gls_" },
+                { name: "Whosty", link: "https://www.youtube.com/@WhostyIO" },
+            ],
+            twitch: [
+                { name: "awmsgaming" },
+                { name: "surviv_skiper" },
+            ],
+        },
```
leia-uwu, line 40: imo we shouldn't have any defaults on the default config

should be up to server hosters to chose what youtubers and streamers they want to display

leia-uwu (changes requested):

client/src/siteInfo.ts
lmssiehdev: tbh fetching on page load is fine

server/src/api/apiServer.ts
```diff
@@ -63,23 +64,32 @@ export class ApiServer {
     clientTheme = Config.clientTheme;
 
     captchaEnabled = Config.captchaEnabled;
+    youtubeCache: { name: string; link: string }[] = [];
+    twitchCache: {
+        name: string;
+        viewers: number;
+        url: string;
+        img: string;
+    }[] = [];
```
DamienVesper: Use `Array<T>` instead.

server/src/api/apiServer.ts
```diff
@@ -63,23 +64,32 @@ export class ApiServer {
...
 
     getSiteInfo(): SiteInfoRes {
+        const featuredYt = util.randomItem(this.youtubeCache) ?? { name: "", link: "" };
         const data: SiteInfoRes = {
             modes: this.modes,
             pops: {},
-            youtube: { name: "", link: "" },
-            twitch: [],
+            youtube: featuredYt,
+            twitch: this.twitchCache ?? [],
```
DamienVesper: `this.twitchCache` will never be a nullish value, so the right-hand side of the nullish coalescing operator will never be reached.

server/src/api/apiServer.ts
```diff
@@ -111,6 +121,51 @@ export class ApiServer {
...
+        };
+        refresh();
+        setInterval(refresh, 300000);
+    }
+
+    private async fetchTwitchData(name: string) {
+        if (!Config.secrets.TWITCH_CLIENT_ID || !Config.secrets.TWITCH_OAUTH) {
+            return;
+        }
+        try {
```
DamienVesper, line 136: The only thing that can fail in this try-catch is the fetch request, so instead opt to handle the error via [Promise API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) for the `Promise` returned by the fetch request. 

Try-catch should be avoided when possible.

server/package.json
DamienVesper, line 1: This file should not have its indentation changed.

DamienVesper (changes requested):

server/src/api/apiServer.ts
lmssiehdev: `const data = await res.json();` or if the data returned from api changed can cause it to fails as well.

> Try-catch should be avoided when possible.

not when making fetch requests.

server/src/api/apiServer.ts
```diff
@@ -111,6 +121,51 @@ export class ApiServer {
...
     }
+
+    private startFeaturedRefresh() {
+        const refresh = async () => {
+            this.youtubeCache = [...Config.featured.youtubers];
+            const streamers: typeof this.twitchCache = [];
+            for (const s of Config.featured.twitch) {
+                const twitchData = await this.fetchTwitchData(s.name);
+                if (twitchData) streamers.push(twitchData);
+            }
```
lmssiehdev: we display at most two streamers in the UI I think, so we should avoid fetching data for all the streamers

I wonder if it's possible to fetch all the data in one request, can we pass more users here? if not you should refactor this into a `Promise.allsettled`.

`https://api.twitch.tv/helix/streams?user_login=${name}`

server/src/api/apiServer.ts
```diff
@@ -111,6 +116,49 @@ export class ApiServer {
...
+        };
+        refresh();
+        setInterval(refresh, 300000);
+    }
+
+    private async fetchTwitchData() {
+        if (!Config.secrets.TWITCH_CLIENT_ID || !Config.secrets.TWITCH_OAUTH) {
+            return [];
+        }
+        const streamers = Config.featured.streamers.slice(0, 3);
```
lmssiehdev, line 132: we will probably want to shuffle this instead of getting the first three streamers, but let's worry about that after we get 3 people to stream survev first :clueless:

DamienVesper:
Closing as stale.

PR #279, opened 2025-10-13
Bunch of fixes for game server <-> api server requests

leia-uwu:
makes saving games be way more reliable :)

also some random changes like making game server not request proxy check, only API does now, saving network requests

### DIFF
--- a/server/src/api/routes/private/private.ts
+++ b/server/src/api/routes/private/private.ts
@@ -1,4 +1,4 @@
-import { and, eq } from "drizzle-orm";
+import { and, eq, inArray } from "drizzle-orm";
 import { Hono } from "hono";
 import { z } from "zod";
 import { saveConfig } from "../../../../../config";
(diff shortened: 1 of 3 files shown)


server/src/api/routes/private/private.ts
```diff
@@ -118,6 +119,26 @@ export const PrivateRouter = new Hono<Context>()
...
 
+        const gameIds = [...new Set(data.matchData.map((d) => d.gameId))];
+
+        // i really don't want the game server to insert duplicated games by accident
+        // when saving lost game data...
+        const exists = await db
+            .select({
+                gameId: sql`DISTINCT(${matchDataTable.gameId})`,
+            })
+            .from(matchDataTable)
```
lmssiehdev, line 130: why attempt to remove duplicates from matchData, doesn't the way we're calling /save_game guarantees they all belong to the same game?

btw drizzle provides a selectDistinct helper which is much cleaner https://orm.drizzle.team/docs/select#distinct-select

also not sure 100% sure but it's probably possible to avoid that exist check with
```js
await db.insert(matchDataTable)
.values(matchData)
.onConflictDoNothing()
```
this would require adding a new composite uniqueIndex tho, so not worth it in this case

leia-uwu: > doesn't the way we're calling /save_game guarantees they all belong to the same game?

im too paranoid some weird condition will make that not true :sob:

lmssiehdev: i gotta say this is the most useless guard i seen in the codebase so far

leia-uwu: ok but what if the game server dies after sending the games but before deleting them from the local files so when it reboots it tries to send them again!!!
the chance of it happening is as high as of a 10km asteroid hitting earth tomorrow but it can happen!!!!!

lmssiehdev: > const gameIds = [...new Set(data.matchData.map((d) => d.gameId))]

the existing db check is fine, but i do wonder what's the chance of having two different gameIds here :clueless:

leia-uwu: > but i do wonder what's the chance of having two different gameIds here :clueless:

because of the new code that retries to save lost games every hour, if theres more than 1 lost game it will send all of them in the same request

lmssiehdev: ah, that answers my original question. i see i see.

PR #282, opened 2025-10-23
Cobalt changes

singe725:


shared/defs/maps/baseDefs.ts
```diff
@@ -397,18 +397,23 @@ export const Main: MapDef = {
...
         ],
         tier_guns_common_healer: [
             { name: "mk12", count: 1, weight: 1 },
             { name: "m39", count: 1, weight: 1 },
+            { name: "vss", count: 1, weight: 1 },
         ],
         tier_guns_common_demo: [
             { name: "m870", count: 1, weight: 1 },
             { name: "spas12", count: 1, weight: 0.5 },
+            { name: "spas12", count: 1, weight: 1 },
```
DamienVesper: duplicate key spas12

DamienVesper (changes requested):
x

shared/defs/maps/baseDefs.ts
lmssiehdev: how do u even catch this

singe725: By... looking at the file and comparing it to the changes we agreed on

DamienVesper: > how do u even catch this

u forget i have a built-in mental linter

PR #284, opened 2025-10-23
enhance: updated gun models

The-Overseer1:
New gun models for:
- Mosin
- Vector (9mm)
- Vector (.45 ACP)
- Peacemaker
- Model 94
- BLR
- Scar-H
- Scout Elite
- M1100


InkyDarkBird:
damn

shared/defs/gameObjects/gunDefs.ts
DamienVesper, line 1: Use hexadecimal values for the tints instead of base 10.

DamienVesper (changes requested):
See requested changes.

Additionally, I feel that you can reduce the file size a bit more, but I am not an SVG expert, so perhaps ask someone else who has submitted SVGs like eipi or Leia as to what other bloat to remove.

shared/defs/gameObjects/gunDefs.ts
The-Overseer1: Done. File size is good enough.

PR #288, opened 2025-10-29
refactor: bun monorepo

lmssiehdev:
I've been spoiled with turborepo recently, and I've been wanting to learn how it works under the hood for a while now, this is my attempt at demystifying that.

this is PR is a refactor to use bun workspaces, and the apps/ packages/ convention which looks like it's becoming a standard

(the api/game server runtime is still on node)

```md
├── apps
│   ├── client (bun)
│   ├── bot (bun)
│   ├── api (node)
│   └── server (node)
├── packages
│   ├── shared
│   ├── types
│   ├── lib
│   └── config (blank, implement it yourself if you want to familiarize yourself with the pr)
```

dropped esbuild as a dependency in favor of https://bunup.dev which uses bun under the hood.

also switched to using rolldown for the vite build, cuz why not? Now the client builds in 1s instead of 7s for me.

there is nothing interesting about the PR besides how it avoids building the packages in development and uses the ts files directly. I really didn't want to compromise on the speed and this took a LOT and I mean a LOT of time to get right, and I'm happy with how it turned out.

it's all possible thanks to https://www.typescriptlang.org/tsconfig/#customConditions, zod and other libraries use a similar approach.

```json
  // in the ts config
  "compilerOptions": {
    "customConditions": ["@survev/source"]
  }
```

```json
  // in pacakges/lib package.json
  "name": "@survev/lib",
  "exports": {
    "./net": {
      "import": {
        "@survev/source": "./src/net/index.ts",
        "default": "./dist/index.js"
      }
    },
  }
```

```json
  // in apps/server package.json
  "dependencies": {
    "@survev/lib": "workspace:*",
  }
```

```js
  // in server/src/game.ts
  // pass the same custom condition define in the tsconfig "tsx watch --conditions=@survev/source"
  import * as net from "@survev/lib/net";
```

I don't think this should be merged though. monorepos are hard and there is a lot of nuance to them and the previous setup works just fine.

```json
{
  // lot of hacks 🤮
  "db:generate": "bun --conditions=@survev/source node_modules/.bin/drizzle-kit generate --config=drizzle.config.ts",
}
```


lmssiehdev:
it's in functional state, i'm happy, closing.

PR #289, opened 2025-10-29
Disable Auto-Pickup for Trick or Treat on Mobile

InkyDarkBird:
I don't know if this code fails in any hypothetical scenarios and I don't know how to code a new function that checks if the player has a Trick-or-Treat perk.

(I'm pretty sure hasPerk() doesn't work unless I want to spam a bunch of or statements)

i hope this code works

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -1942,12 +1942,9 @@ export class Player extends BaseGameObject {
                         break;
                     }
                     case "perk": {
-                        if (!this.perks.find((perk) => perk.droppable)) {
-                            if(this.perks.find(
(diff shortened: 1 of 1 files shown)


InkyDarkBird:
https://github.com/user-attachments/assets/0bb973ed-9bf7-4a69-9207-b973be6c2362

singe725:
> was this discussed with the devs? if yes, then i can take a look.

Not directly but I'd classify it as a bug since mobile players shouldn't be forced to pick up trick or treat perks.

InkyDarkBird:
wait im not sure if i have the most updated version lemme check again

PR #291, opened 2025-11-01
Fix: Turkey Mode Spawns, Pass 1 Outfit Drop On Death, Changelog

NotASwifty13:
-Put woods Mosin cache into Turkey Mode to fit the mode theme better. 
-Removed berry bush caches and replaced with leaf pile caches (now a more reasonable mix).
-Made Survivr Pass 1 outfits drop on death by commenting noDropOnDeath: true out. We'll need to uncomment those with the pass release. 
-Fixed "Xth" date in changelog.

### DIFF
--- a/balance.txt
+++ b/balance.txt
@@ -34,34 +34,34 @@ VSS:
 + changed potato quality from 1 (rare) to 0 (common)
 
 Flak Jacket:
-Size increase +0.2 -> +0.1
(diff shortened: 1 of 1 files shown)


DamienVesper (changes requested):
Update `balance.txt` to reflect cache changes.

lmssiehdev:
image

never forget.

PR #292, opened 2025-11-01
Fix: Halloween/Shared Atlas Conflict

NotASwifty13:
-Removed tree_08 and tree_07 from halloween.ts since Hex put them in shared and shared is included in every map's assets. Should resolve the bug where trees don't spawn in Turkey after playing a game in Halloween.

### DIFF
--- a/tests/src/maps.test.ts
+++ b/tests/src/maps.test.ts
@@ -1,6 +1,6 @@
 import "./testHelpers";
 import { describe, expect, test } from "vitest";
-
+import { Atlases } from "../../client/atlas-builder/atlasDefs";
 import { type MapDef, MapDefs } from "../../shared/defs/mapDefs";
 import { Constants } from "../../shared/net/net";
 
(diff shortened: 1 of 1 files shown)


DamienVesper (changes requested):
The entirety of https://github.com/survev/survev/commit/31777f746c354bb5974b73d6250f5fc28ab0c6bd is a mistake. These assets should not be be in shared. They should be included as-needed in the necessary atlases, NOT in shared. Any asset that is ever included in an atlas SHOULD have a non-zero probability to be used according to the map definitions.

If you are going to make a PR regarding these atlases, please fix that issue instead of supporting its mistakes.

NotASwifty13:
> The entirety of [31777f7](https://github.com/survev/survev/commit/31777f746c354bb5974b73d6250f5fc28ab0c6bd) is a mistake. These assets should not be be in shared. They should be included as-needed in the necessary atlases, NOT in shared. Any asset that is ever included in an atlas SHOULD have a non-zero probability to be used according to the map definitions.
> 
> If you are going to make a PR regarding these atlases, please fix that issue instead of supporting its mistakes.

Alright, so nothing should've been removed from Woods or Halloween, and Woods should've been included in Turkey? Turkey doesn't currently have its own atlas, I wouldn't think it would need one.

DamienVesper:
On the other hand, Leia had proposed an autumn atlas instead in ce3cdbaa4e47fee0ce53ac448d9101cb145ba766, so that could be a potential alternative.

PR #274, opened 2025-10-02
feat: autumn (turkey) mode

Hexevillius:
note: check if the atlases are correct (and reasonable)

PR #293, opened 2025-11-07
feat: new 'The Pro' skin unlock for players with 500+ wins

NAMERIO:
New skin for a player with over 500 wins.

### DIFF
--- a/server/src/api/db/schema.ts
+++ b/server/src/api/db/schema.ts
@@ -64,7 +64,7 @@ export const itemsTable = pgTable(
         source: text("source").notNull().default("unlock_new_account"),
         status: integer("status").notNull().default(ItemStatus.New),
     },
-    (table) => [primaryKey({ columns: [table.userId, table.source] })],
+    (table) => [primaryKey({ columns: [table.userId, table.type] })],
 );
 
 export const matchDataTable = pgTable(


leia-uwu:
this code sucks

ei-pi:
lul

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
             tint: 0xffffff,
         },
     }),
+    outfitThePro: defineOutfitSkin("outfitBase", {
+        name: "The Pro",
+        noDropOnDeath: true,
+        rarity: Rarity.Common,
+        lore: "The Pros can wear bright colors, they have no need to hide. You Do.",
```
DamienVesper: This description could be improved. It doesn't use bright colors, also the grammar is wrong.

shared/defs/gameObjects/unlockDefs.ts
```diff
@@ -388,4 +389,10 @@ export const UnlockDefs: Record<UnlockDefKey, UnlockDef> = {
         free: true,
         unlocks: ["outfitDarkShirt"],
     },
+    unlock_500_wins: {
+        type: "unlock",
+        name: "Champion Skin",
```
DamienVesper: Why is this attributed as "Champion Skin" while being named "The Pro?" Typically you would want to write the reason for acquirement here. Consider changing this as well.

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +143,29 @@ export const PrivateRouter = new Hono<Context>()
...
+        const items = playerWins.filter(({wins}) => wins >= 500)
+            .map(u => ({
+                    userId: u.userId as string,
+                    source: "unlock_500_wins",
+                    type: "outfitThePro",
+                    timeAcquired: Date.now(),
+                }));
+
+        await db
+            .insert(itemsTable)
```
DamienVesper: Once a player reaches 500 wins, every game that this player proceeds to join will result in them obtaining a new copy of the skin. Duplicated items have the additional feature of crashing the client.

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +143,29 @@ export const PrivateRouter = new Hono<Context>()
 
         await db.insert(matchDataTable).values(matchData);
         await logPlayerIPs(matchData);
+        const playerIds = matchData.map(d => d.userId).filter(d => d != null);
```
DamienVesper: This should filter for players that won, to avoid unnecessary queries.

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +143,29 @@ export const PrivateRouter = new Hono<Context>()
...
+        const playerWins = await db
+
+            .select({
```
DamienVesper, line 158: Does not filter for the same player being in the game twice (i.e. joining a squad as yourself twice).

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +143,29 @@ export const PrivateRouter = new Hono<Context>()
...
+                userId: matchDataTable.userId,
+                wins: sql<number>`SUM(CASE WHEN ${matchDataTable.rank} = 1 THEN 1 ELSE 0 END)`.as("wins"),
+            })
+            .from(matchDataTable)
+            .where(inArray(matchDataTable.userId, playerIds))
+            .groupBy(matchDataTable.userId);
+
+        const items = playerWins.filter(({wins}) => wins >= 500)
```
DamienVesper: This can be written as part of the database query for a non-trivial improvement in performance.

shared/defs/gameObjects/unlockDefs.ts
```diff
@@ -388,4 +389,10 @@ export const UnlockDefs: Record<UnlockDefKey, UnlockDef> = {
...
+        free: true,
+        unlocks: ["outfitThePro"]
```
leia-uwu: this unlock is set here but then on the code to actually give the skin its hardcoded to `outfitThePro` instead of giving items from this unlocks array

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
...
+        skinImg: {
+            baseTint: 16303476,
```
leia-uwu: should be in hex not decimal

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
...
+            baseSprite: "player-base-01.img",
+            handTint: 3300450,
```
leia-uwu: should be in hex not decimal

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
...
+            handSprite: "player-hands-01.img",
+            footTint: 3300450,
```
leia-uwu: should be in hex not decimal

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
...
+            footSprite: "player-feet-01.img",
+            backpackTint: 2508377,
```
leia-uwu: should be in hex not decimal

client/atlas-builder/defs/loadout.ts
```diff
@@ -421,6 +421,7 @@ export const LoadoutAtlas: AtlasDef = {
         "loot/loot-shirt-outfitCobaltShell.svg",
         "loot/loot-shirt-outfitDarkGloves.svg",
         "loot/loot-shirt-outfitDarkShirt.svg",
+        "loot/loot-shirt-outfitThePro.svg",
```
leia-uwu: sort them alphabetically smh

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
             tint: 0xffffff,
         },
     }),
+    outfitThePro: defineOutfitSkin("outfitBase", {
+        name: "The Pro",
+        noDropOnDeath: true,
+        rarity: Rarity.Common,
```
leia-uwu: Should be a higher rarity

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +143,29 @@ export const PrivateRouter = new Hono<Context>()
...
+                    userId: u.userId as string,
+                    source: "unlock_500_wins",
+                    type: "outfitThePro",
+                    timeAcquired: Date.now(),
+                }));
+
+        await db
+            .insert(itemsTable)
+            .values(items)
+            .onConflictDoNothing()
```
leia-uwu: missing semicolon, please lint

server/src/api/routes/private/private.ts
lmssiehdev: Note the `.onConflictDoNothing()` so this is trying to upsert and ignore on duplicates

I'll have to recheck the docs but I think it would require a target passed into it to work properly.

Unrelated, but we should probably add a composite primary key on the Id and the source in the items table, to prevent duplicates from getting inserted by mistake

And we should filter them when fetching the profile just to be safe

(On phone, excuse typos)

server/src/api/routes/private/private.ts
lmssiehdev: This is done to give the skin to people who had 500 wins before this update without requiring them to win to get it

server/src/api/routes/private/private.ts
lmssiehdev: I was too lazy to figure out the syntax for that

lmssiehdev:
I was asked to implement this great idea, I refused

I knew namerio would be a up for the job

This PR is great. So great Leia got involved.

server/src/api/routes/private/private.ts
```diff
@@ -143,29 +144,59 @@ export const PrivateRouter = new Hono<Context>()
...
+                sql`SUM(CASE WHEN ${matchDataTable.rank} = 1 THEN 1 ELSE 0 END) >= 500`,
+            );
+
+        const existing = await db
+            .select({ userId: itemsTable.userId })
+            .from(itemsTable)
+            .where(
+                and(
+                    inArray(itemsTable.userId, playerIds),
+                    eq(itemsTable.type, "outfitThePro"),
```
leia-uwu: thats so funny you fixed it in one place but at the same time did the same mistake again a few lines above it

server/src/api/routes/private/private.ts
leia-uwu: that seems really wasteful 
could just give it manually once the update is out to avoid doing unnecessary DB queries every single time a game is saved, from looking at the database only around 700 accounts have over 500 wins currently lol, writing a script to do that is not hard, and even then having people have to win an extra game to get the skin even if they already have over 500 is not that big of a deal... if you got 500 wins you can surely get another one lol

from looking at the database around 30% of players are logged in
without filtering for players that won that will query on average 24 players on every 80 players lobby, instead of only between 1 to 4 players depending on the team mode
(speaking of that we should probably blacklist 50v50 when counting wins for obvious reasons: too easy to win on that mode, it counts as win even if you died 10 seconds after joining but your team won, and also it will check for up to 50 players on each saved game)

server/src/api/routes/private/private.ts
leia-uwu: > Unrelated, but we should probably add a composite primary key on the Id and the source in the items table, to prevent duplicates from getting inserted by mistake

yeah, can't we just make a key thats literally `${userId}-${itemId}` lol

server/src/api/routes/private/private.ts
lmssiehdev: > that seems really wasteful

I'm well aware, and I agree just manually giving the skin will help remove all the rough edges around this code, the thing is I didn't believe this will get merged, and I didn't know you were are around still to give the prev winners the skin

I was harassed over the vc, so i threw something together for them quickly with zero optimizations done lol, couldn't be arsed.

shared/defs/gameObjects/outfitDefs.ts
lmssiehdev: I wonder how hard it is to write a biome rule for this

server/src/api/routes/private/private.ts
lmssiehdev: > yeah, can't we just make a key thats literally ${userId}-${itemId} lol

that would work yea, but feels hacky

composite keys are more versatile and maybe needed in other use cases beside this, plus they are indexed by default, so there is perf win there.

talking about perf wins, i should really add an index to the session table, been planning to for months :sob:

server/src/api/routes/private/private.ts
lmssiehdev: > matchData.map(d => d.userId).filter(d => d != null)

mapping before a filter.. damien is really blind

shared/defs/gameObjects/outfitDefs.ts
lmssiehdev: got something working lol, need some optimizations since it's slow af

should we auto fix these? i'm worried about false positives


image

lmssiehdev (changes requested):
few nitpicks, looks good beside that, good work.

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +144,59 @@ export const PrivateRouter = new Hono<Context>()
...
+                unlock.unlocks.map((type) => ({
+                    userId: userId!,
+                    source: "unlock_500_wins",
+                    type,
+                    timeAcquired: Date.now(),
+                })),
+            );
+
+        if (items.length > 0) {
+            await db.insert(itemsTable).values(items);
```
lmssiehdev: you should keep the `.onConflictDoNothing()` and add a composite key to prevent duplicates from being inserted

```js
// schema.ts itemsTable
    status: integer("status").notNull().default(ItemStatus.New),
}, (table) => [
    primaryKey({ columns: [table.userId, table.source] }),
])
```

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +144,59 @@ export const PrivateRouter = new Hono<Context>()
...
+                    inArray(itemsTable.userId, playerIds),
+                    inArray(itemsTable.type, unlockItems),
+                ),
+            );
+
+        const existingSet = new Set(existing.map((e) => e.userId));
+
+        const items = playerWins
+            .filter(
+                ({ wins, userId }) => wins >= 500 && userId && !existingSet.has(userId),
```
lmssiehdev: you can remove the existingSet by using `.selectDistinct({ userId: itemsTable.userId })` to directly return distinct players

you should also use  `notInArray(itemsTable.type, unlockItems)` to only select users that don't have the skin

`({ wins, userId }) => wins >= 500 && userId && !existingSet.has(userId)` can be removed all together since the playerWins is only returning players that have enough wins

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +144,59 @@ export const PrivateRouter = new Hono<Context>()
...
+            .select({
+                userId: matchDataTable.userId,
+                wins: sql<number>`SUM(CASE WHEN ${matchDataTable.rank} = 1 THEN 1 ELSE 0 END)`.as(
+                    "wins",
+                ),
+            })
+            .from(matchDataTable)
+            .where(inArray(matchDataTable.userId, playerIds))
+            .groupBy(matchDataTable.userId)
+            .having(
```
lmssiehdev, line 167: no need to recalculate the sum here since you are using the `.as("wins")` alias, `.having(sql`wins >= 500`)` would do.

this and the existing check can probably be done in one query, but I don't think there will be much to gain from that, so two queries are fine for now

still you should move this code out of the `/save_game` endpoint into it's own function

shared/defs/gameObjects/outfitDefs.ts
```diff
@@ -842,6 +842,26 @@ const SkinDefs: Record<string, OutfitDef> = {
...
+            handTint: 0x325c62,
+            handSprite: "player-hands-01.img",
+            footTint: 0x325c62,
+            footSprite: "player-feet-01.img",
+            backpackTint: 0x274659,
+            backpackSprite: "player-circle-base-01.img",
+        },
+        lootImg: {
+            sprite: "loot-shirt-outfitThePro.img",
+            tint: 0xffffff,
```
lmssiehdev, line 862: any reason for adding a new svg instead of using the "loot-shirt-01.img" with a custom tint?

NAMERIO: Other skins like the semi pro and similar items each have their own SVG file too, so I followed the same approach

server/src/api/routes/private/private.ts
leia-uwu: `table.source`? shouldn't it be `table.type` instead?

shared/defs/gameObjects/outfitDefs.ts
leia-uwu: > should we auto fix these? i'm worried about false positives

maybe have then as an unsafe auto fix? so they are only fixed if you pas `--unsafe` when running biome

server/src/api/routes/private/private.ts
lmssiehdev: correct

shared/defs/gameObjects/outfitDefs.ts
lmssiehdev: looks like it's not possible yet https://github.com/biomejs/biome/issues/5687

server/src/api/routes/private/private.ts
DamienVesper: There's a reason you don't have merge perms to the main codebase any longer, preacher. Such is one of the reasons.
Even if Leia wasn't around, if someone wants the skin, winning 1 game is not such a difficult concept to achieve. There is no reasonable justification for tanking performance for this.

Either way, that implementation had no place to exist in its previous form.

server/src/api/routes/private/private.ts
DamienVesper: Technically according to big O-notation, both ways are still O(n)...so I'm not incorrect for ~~missing~~ not mentioning it...
But yes, I agree.

server/src/api/routes/private/private.ts
```diff
@@ -143,6 +144,42 @@ export const PrivateRouter = new Hono<Context>()
 
         await db.insert(matchDataTable).values(matchData);
         await logPlayerIPs(matchData);
+        const unlock = UnlockDefs["unlock_500_wins"];
+        const unlockItems = unlock.unlocks;
+        const playerIds: string[] = [
+            ...new Set(
```
DamienVesper, line 150: Creating a set just to use a spread operator is completely unnecessary.
Also am not a fan of filtering and mapping, maybe use reduce or flatMap instead.

server/src/api/routes/private/private.ts
DamienVesper: > Last time I checked, I didn't have merge perms because I said I would rather not have them after you gave them to me manually cuz you fucked up the perms.

I removed merge perms for you intentionally with the CODEOWNERS file in 6994f6a6e616f293448b95a986a2429700898d27. I specifically only gave you merge perms for PRs modifying the API server. Given that I don't like touching this project, I walked back my stance. But seeing how you've began to review PRs without caring to look at them, I think my initial standing was the correct position to take.

Do not argue you didn't merge PRs without looking at them: #274.
Code quality is not the end all, be all, but there is a minimum standard that was not met.

> No one is as petty as you, to use something that I wrote in the discord textarea in two minutes because namerio wanted some starting code to justify that.

If you don't want to suggest some code because it is poor, just say no. According to your own wording:

> I was harassed over the vc, so i threw something together for them quickly with zero optimizations done lol, couldn't be arsed.

Nowhere in this statement do you mention that such code should not be used. By writing this code AND replying in such a manner that does not acknowledge the fault of such code, it seems you are defending it on the basis of laziness, which again brings us back to the point of a minimum code quality threshold that is not being met.

You are also not obligated to fulfill all requests. Therefore, harassment is not a justification for poor code.

> It was never the final code, nor was it intended to be pushed like that without modifications.

The PR was made with such code. I review it as such. If you defend such code, I review **you** as such.

> I was against this idea of this PR then, and still against it now, they should wait for the survey pass if they want something like this.

I suppose we will close the PR then.

> I gave my very valid justification and reasonable justification above, I wasn't interested in opening my IDE for this PR.

Ignorance is not an excuse to avoid meeting expectations. Laws cannot be broken on the justification of ignorance.

In this instance, we are talking about your defense of poor code, not laws. Yet still, you are attempting to circumvent simple optimizations on the basis of ignorance. Thus, the same logic applies.

> Like I said when responding to Leia, the original code rough edges can be removed the moment that someone with access to the api server is around
and I already suggested some performance changes in this PR itself

Once again, attempting to justify poor code.

> If I thought you were sensible enough or that your opinion mattered, I would have referred you to look at my accounts PR

Unfortunately, I control merging now. My opinion, sadly, does matter.

To bypass such, you would have to either contact Leia or Mistik. Leia not only lacks sympathy for such trivialities but also would not in some capacity agree with this PR. Mistik controls the production repository, and if you would like to deal with that, be my guest. I believe that it is a sufficient price to pay in the downcast of your emotions following a brief run-through of the commit history of that repository.

> So by tanking the performance, you are referring to at most 20ms every time a game ends

I would like to see the % change included in this metric. Stating an absolute change of 20ms does not hold much value without relating it to the current implementation.

Additionally, there is still no reason to unnecessarily calculate data that will not be used. This is not a USACO Pails problem where it is faster to bruteforce the solution, rather than be smart about it.

> It still needs to be improved and as opposed to what you were trying to claim, I wouldn't have merged it but let's not be dramatic with the language

I understand you are attacking my diction. For it is a matter of language, I will allow it.

> It will just come across as ignorant, and it will feed into my bias of you having no clue what you are talking about

It is possible I do not know what I am talking about. Then again, logic prevails and you are at least partially agreeing with my code analysis. Therefore, I have some idea of what I am talking about, rather than no idea.

DamienVesper:
Closing as blocked by Pass 1 implementation.

PR #297, opened 2025-11-15
feat: Auto random perk assigning

bibizu:
Adds a couple of properties to the mapDef.ts that allows for players to be assigned random perks globally from a list of perks.

Adds a new sample mode "Chaos" that showcases the feature. Ultimately, this mode can be deleted--the first feature can be used to enhance the mode rotation slightly every now and then, such as giving everyone turkey shoot at the start of the game for potato or woods without messing with loot tables.

Also adds in the 3 missing overpressure perks (without images or translations, sorry. I am not an artist nor a multilingual.)

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -29,7 +29,7 @@ export function getConfig(isProduction: boolean, dir: string) {
         regions: {},
         proxies: {},
         modes: [
-            { mapName: "chaos", teamMode: TeamMode.Solo, enabled: true },
+            { mapName: "main", teamMode: TeamMode.Solo, enabled: true },
             { mapName: "main", teamMode: TeamMode.Duo, enabled: true },
             { mapName: "main", teamMode: TeamMode.Squad, enabled: true },
         ],


package.json
```diff
@@ -30,7 +30,8 @@
   "dependencies": {
     "enquirer": "^2.4.1",
     "hjson": "^3.2.2",
-    "tsx": "^4.20.6"
+    "tsx": "^4.20.6",
+    "vite": "^7.2.2"
```
DamienVesper, line 34: Vite does not need to be in the root package.json

package-lock.json
DamienVesper, line 1: We use pnpm as the package manager, and thus, `pnpm-lock.yaml`. You should not have used `npm` or committed this file.

server/src/game/map.ts
```diff
@@ -354,6 +357,31 @@ export class GameMap {
                 );
             });
 
+
+        let totalWeight = 0.0;
+        let perkTable = this.mapDef.gameMode.autoPerkTable
+        for (let i = 0; i < perkTable.length; i++) {
+            totalWeight += perkTable[i].weight;
+        }
+        for (let i = 0; i < <number>this.mapDef.gameMode.autoPerkCount; i++) {
```
DamienVesper, line 366: I don't know why there is a type assertion here.

DamienVesper (commented):
Despite the PR being closed, I will add just a few nitpicking comments not entirely related to the implementation.

The reason the tests failed is due to you using npm and not pnpm (although you shouldn't have made the changes you did to root `package.json` anyway).
Also, please lint your code next time.

There are other issues with the code but it's not worth my time to detail as this was decided not to be implemented. You are free to implement this in a fork though! I'm sure someone would like it in their codebase. We just don't think it belongs in the base game.

bibizu:
Appreciate the review! This was my first time figuring things out in typescript, and also my first pull request to a public repo. I think I must've accidentally also added npm, as I am also using pnpm.

PR #298, opened 2025-11-23
Refactor: Improved Wall Generation 

NotASwifty13:
Been working on this for a while, and have taken week-long breaks in between, so hopefully I haven't left anything in that isn't needed. Looked alright to me. 

Submitting as a request for critique, and overall review of necessity/implementation. Not ready to merge, though everything works perfectly, as far as I know. Let me know if there's anything I need to test, fix, or otherwise edit, let me know.

Barn is completed as a working example. If this system isn't worth the time, I'll scrap it, but I personally don't want to have to make a new wall def for every wall I want to add from here on out, and spawn it on center (or have to always keep re-usability in mind while drawing new buildings).

### DIFF
--- a/shared/defs/mapObjectDefs.ts
+++ b/shared/defs/mapObjectDefs.ts
@@ -1840,6 +1840,29 @@ function createBank<T extends BuildingDef>(e: Partial<T>): T {
                 },
             ],
         },
+        walls: [
+            {
+                type: "brick_wall_base",
+                collision: [
+                    collider.createAabb(v2.create(-25.5, 16.5), v2.create(-2.5, 17.5)),     // brick_wall_ext_23 at (-14, 17) ori:1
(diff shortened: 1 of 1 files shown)


leia-uwu:
initial review: i don't get why you created a custom AABB interface (with inlined type everywhere and string for `type`??) instead of just using `AABB` from coldet.ts

also for positions dont use `write/readFloat` with hardcoded boundaries, use `write/readMapPos`, realistically an obstacle hitbox will never be negative i think, but if thats the case still use `write/readVec` and don't hardcode the 1024 bounds

leia-uwu:
> Do you think the idea is solid overall? I've since done some additional efficiency improvements to make the system even simpler/less lines.

i don't like the idea of sending texture strings in packets
i think we should keep this system only for the invisible walls with no texture and keep the regular breakable walls with textures as regular obstacles, and invisible walls make the most part on big buildings anyway i think

but other than that i think it could work

PR #301, opened 2025-12-01
feat: RIP Rounds perk + Lone Survevr Buffs

InkyDarkBird:
This perk is intended to nearly instantly kill downed players. This synergizes extremely well with Lone Survevr with Takedown.

I coded this because both Swifty and Stu liked it.

image

https://github.com/user-attachments/assets/60f62be2-8081-43c3-8c95-c161213f5371

TODO:
Make the RIP Rounds perk description work properly (I have no idea how to do this)
Scale the image size better and maybe remake the image?

### DIFF
--- a/shared/defs/gameObjects/roleDefs.ts
+++ b/shared/defs/gameObjects/roleDefs.ts
@@ -396,8 +396,8 @@ export const RoleDefs: Record<string, RoleDef> = {
                         },
                         teamcolor,
                     ),
-                { type: "", ammo: 0 },
-                { type: "mirv", ammo: 8 },
+                { type: "pan", ammo: 0 },
+                { type: "", ammo: 8 },
             ],
             backpack: "backpack03",
             helmet: (teamcolor: TeamColor) =>
@@ -411,6 +411,8 @@ export const RoleDefs: Record<string, RoleDef> = {
             chest: "chest04",
             inventory: {
                 mirv: 8,
+                frag: 12,
+                strobe: 3,
                 "8xscope": 1,
                 bandage: 10,
                 healthkit: 1,


singe725:
You can make the perk text work properly by using translations. The image is terrible and not really in line with the rest of the perks, but that's a quick fix. Besides that, this will require a balancing discussion, which we aren't likely to engage in for a bit. It's interesting, but there's zero room for counterplay and it seems very powerful compared to the other lone survivr perks currently.

server/src/game/objects/player.ts
NotASwifty13, line 1: What do you mean perform the Lone Survevr check?

NotASwifty13, line 1: Just some initial glance over thoughts..... I think a much better solution to this perk applying a multiplier to down targets so less ammo is wasted finishing them off.

It'd be better to follow the patterns of ap_rounds, splinter, etc. and use perkProperties to add a multiplier for downed targets. For the record, I thought the perk was interesting, but not in love with it exactly. Would love to see it improved, but I'm not a fan of its current state.

shared/defs/gameObjects/roleDefs.ts
```diff
@@ -370,6 +370,7 @@ export const RoleDefs: Record<string, RoleDef> = {
                 util.weightedRandom([
                     { type: "ap_rounds", weight: 1 },
                     { type: "splinter", weight: 1 },
+                    { type: "rip_rounds", weight: 1 },
```
NotASwifty13, line 373: Maybe this should be a rarer one in the windwalk/combat medic tier.... Not having a perk that does increased damage to alive, shooting threats would really suck. Then stacking of multipliers/or whatever the perk ends up doing might be too OP.

Hence why I have issues with this on lones.

InkyDarkBird: > Maybe this should be a rarer one in the windwalk/combat medic tier.... Not having a perk that does increased damage to alive, shooting threats would really suck. Then stacking of multipliers/or whatever the perk ends up doing might be too OP.
> 
> Hence why I have issues with this on lones.

Should RIP Rounds fully replace Windwalk then? I currently believe Windwalk is the worst perk on the Lone Survevr kit.

InkyDarkBird:
+12 frags
+3 strobes
+Pan

also eliminated MIRVs from being repeated

DamienVesper:
Terrible idea, most random balance changes, incomplete implementation, breaks meta, so not merging.

PR #309, opened 2025-12-21
Game Moderatr suit -> noDrop: true

Kisakay:
I made this change cause i think that this modification is important.
First of all: when playing with mod suit, if you were missclicking with an another stuff and then random player stole this mod suite, this will make confusion into the game itself. An scenario of an cheater (for example) with the moderator suite will be chaotic. And second one, i'm think that, when game moderator start to use the suite, they have to assume the skin for the entire game.


leia-uwu:
inconsistent with other skins, we should do a better fix for loadout skins in general, like maybe only allowing the skin owner to pick it up from the ground

PR #312, opened 2025-12-29
feat: Snow Mode + Gun Models

The-Overseer1:
image_2025-12-29_110804065

Anyways, there's a bunch of fixes and additions to snow mode, as well as 3 new gun models.

### DIFF
--- a/client/public/l10n/de.json
+++ b/client/public/l10n/de.json
@@ -455,6 +455,7 @@
   "game-hud-crowbar_recon": "Brecheisen",
   "game-stonehammer": "Steinhammer",
   "game-sledgehammer": "Vorschlaghammer",
+  "game-iceaxe": "Eispickel",
   "game-hook": "Haken",
   "game-pan": "Pfanne",
   "game-crowbar_scout": "Späher-Brecheisen",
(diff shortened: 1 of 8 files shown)


shared/defs/maps/snowDefs.ts
```diff
@@ -109,11 +214,46 @@ const mapDef: PartialMapDef = {
...
+                warehouse_complex_01: 1,
+                chest_01: 1,
+                chest_03: { odds: 0.2 },
+                mil_crate_02: { odds: 0.25 },
+                tree_02: 3,
+                teahouse_complex_01su: {
+                    small: 1,
+                    large: 2,
+                },
+                stone_04: 3, // had to add the whole table to change this...might want to consider adding some form of partial override - Stu
```
leia-uwu: why put `stone_04` here and then put `stone_04: "stone_04x"` on `spawnReplacements`? could just put `stone_04x` here directly

same for barns, containers, teahouses, etc

shared/defs/mapObjectDefs.ts
```diff
@@ -2961,7 +2961,7 @@ function createBarnBasement<T extends BuildingDef>(e: Partial<T>): T {
                 ori: 0,
             },
             {
-                type: "loot_tier_sledgehammer",
+                type: e.floor_loot || "loot_tier_sledgehammer",
```
leia-uwu: could have just set `loot_tier_sledgehammer` on snow mode to drop the ice axe instead of sledgehammer, no need for a secondary loot tier

PR #318, opened 2026-01-10
feat(WIP): gun stats

lmssiehdev:
still need to collect the stats and save them to the db after each game , but I won't do any more work on this until a proper styling is introduced, i'm not gonna bother with modifying the original css and i can't go far with just using the raw style attribute

+1 for https://unocss.dev/ from me

image

### DIFF
--- a/server/src/api/db/schema.ts
+++ b/server/src/api/db/schema.ts
@@ -5,6 +5,7 @@ import {
     integer,
     json,
     pgTable,
+    primaryKey,
     serial,
     text,
     timestamp,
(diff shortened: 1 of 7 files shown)


server/src/api/db/schema.ts
```diff
@@ -46,6 +47,24 @@ export const usersTable = pgTable("users", {
...
+export const weaponStats = pgTable(
+    "weapon_stats",
+    {
+        userId: text("user_id")
+            .notNull()
+            .references(() => usersTable.id, {
+                onDelete: "cascade",
+                onUpdate: "cascade",
+            }),
+        weaponId: text("weapon_id").notNull(),
```
leia-uwu, line 59: weaponType is better imo, keeps consistent with what is used everywhere else in the codebase for definition type strings

shared/types/stats.ts
```diff
@@ -152,3 +152,24 @@ export type LeaderboardResponse = {
...
+export const zWeaponStatsRequest = z.object({
+    slug: z.string(),
+});
+
+export type WeaponStatsParams = z.infer<typeof zWeaponStatsRequest>;
+
+export type WeaponStat = {
+    type: string;
+    name: string;
+    img: string;
```
leia-uwu, line 168: name, img and color should be calculated client side

lmssiehdev: i didn't want to load GameObjectDefs on the /stats/ route, but looks like it's loaded already

since they are loaded already i don't mind, but would rather keep them server side and optimize the bundle

leia-uwu: well name will need to be client side regardless for translations

lmssiehdev: good catch

PR #319, opened 2026-01-10
fix: avoid attempting to read the lost_game_data directory if it doesn't exist

lmssiehdev:
this was spamming the webhook


server/src/gameServer.ts
```diff
@@ -127,6 +129,9 @@ class GameServer {
         const games: SaveGameBody["matchData"] = [];
 
         const dir = path.resolve("lost_game_data");
+
+        if (!existsSync(dir)) return;
```
leia-uwu, line 132: we already import `fs` so use `fs.existsSync` and remove the duplicated import

PR #320, opened 2026-01-10
refactor: split CI

leia-uwu:
using the github ci `paths` list each individual job will only run if a file related to it is modified, so for example client CI won't run when theres only changes to the server, also makes them all parallel since they are individual jobs now

### DIFF
--- a/client/src/account.ts
+++ b/client/src/account.ts
@@ -2,6 +2,7 @@ import $ from "jquery";
 import type {
     LoadoutRequest,
     LoadoutResponse,
+    ProfileResponse,
     RefreshQuestRequest,
     RefreshQuestResponse,
     SetItemStatusRequest,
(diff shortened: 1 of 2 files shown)


.github/workflows/biome.yml
```diff
@@ -0,0 +1,42 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - master
+    paths:
```
lmssiehdev: >mfw no fall through

```YAML
paths:
pull_request:
   ...
```

but they have whatever this is, it will help remove some of the duplication https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases

lmssiehdev:
also mandatory plug for https://www.blacksmith.sh/

PR #322, opened 2026-01-11
feat:  a very crappy functionality for multiple revives at once

InkyDarkBird:
Yes, I know the UI does not update with the revive duration.

https://github.com/user-attachments/assets/322f6ee2-b214-4072-ba19-d6fe499a3b1d

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -819,6 +819,8 @@ export class Player extends BaseGameObject {
     aimLayer = 0;
     dead = false;
     downed = false;
+    reviveProgress = 0;
+    reviveDuration = 8;
 
     downedCount = 0;
     /**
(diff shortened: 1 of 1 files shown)


leia-uwu:
you should extend `tests/src/reviving.test.ts` to include tests with multiple revivers

InkyDarkBird:
I should probably also add checks for the revivers not being in a reviving anymore, but I know that that's going to fail right now so....

leia-uwu:
> Make the revive timer actually sync correctly (I have no idea if this is even possible with the pieTimer being based off of the static Action system)

i think it could work if instead of multiplying the delta time added to the action, you modified the action max timing when a player starts or stops reviving

something like this:

```ts
const actionT = math.clamp(action.time / action.duration, 0, 1);
action.duration = GameConfig.player.reviveDuration * revivingSpeed;
action.time = action.duration * actionT;
```
(don't forget to set `actionDirty = true` on all the players reviving and being revived after this)

PR #335, opened 2026-01-17
feat: break cheats

whymahdi:
modern cheats (forks of surplus) determine which property is which based on counters of datatypes in properties of game objects: https://pastebin.com/raw/REaSt5PV
this should break all cheats for the foreseeable future since surplus is out of commission (deleted it)
if you dont like the signature scrambler thing ms brazillian leia skip it but pls merge the rest thatll do on its own
just me not maintaining surplus anymore will fix things on its own as long as you break it once
thx bye

### DIFF
--- a/client/src/debug/debugHUD.ts
+++ b/client/src/debug/debugHUD.ts
@@ -151,16 +151,16 @@ export class DebugHUD {
                 addCount(name, pool.m_activeCount, pool.m_getPool().length);
             };
 
-            addPool("Players", game.m_playerBarn.playerPool);
-            addPool("Loot", game.m_lootBarn.lootPool);
-            addPool("Projectiles", game.m_projectileBarn.projectilePool);
+            addPool("Players", game.m_playerBarn.m_playerPool);
+            addPool("Loot", game.m_lootBarn.m_lootPool);
+            addPool("Projectiles", game.m_projectileBarn.m_projectilePool);
             addPool("Obstacles", game.m_map.m_obstaclePool);
             addPool("Buildings", game.m_map.m_buildingPool);
             addPool("Structures", game.m_map.m_structurePool);
-            addPool("Decals", game.m_decalBarn.decalPool);
-            addPool("Dead Bodies", game.m_deadBodyBarn.deadBodyPool);
+            addPool("Decals", game.m_decalBarn.m_decalPool);
+            addPool("Dead Bodies", game.m_deadBodyBarn.m_deadBodyPool);
             addPool("Smoke", game.m_smokeBarn.m_smokePool);
-            addPool("Airdrops", game.m_airdropBarn.airdropPool);
+            addPool("Airdrops", game.m_airdropBarn.m_airdropPool);
 
             const activeParticles = game.m_particleBarn.particles.filter(
                 (p) => p.active,
--- a/client/src/emote.ts
+++ b/client/src/emote.ts
@@ -140,8 +140,8 @@ export class EmoteBarn {
     emoteMouseTriggered = false;
     emoteScreenPos = v2.create(0, 0);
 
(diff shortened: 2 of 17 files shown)


whymahdi:
in dms thx

whymahdi:
surplus was the biggest cheat used by the most people, if it breaks without me supporting it anymore it will drastically reduce the amount of cheaters for a really long time thats the fucking point ok

noam0001:
@chickenpoo351 you keep on spewing nonsense, there obviously is no permanent solution to cheating, this is a browser game, there is no universal anticheat.
except there is but they're mostly centered around prototype pollution or in other cases just detecting script tags.

you make it very clear that you know very little about what is being talked about here.

Surplus does not use any asset swapping, code patching, or any swapping of these sorts.
It relies solely on prototype pollution, on a Function.prototype.call hook.

This pr is not meant to stop cheating forever as thats impossible, it would have to be a cat and mouse game where the devs of the game could come out advantageous due to silent delayed ban flags.

There is no case in which ESP for example will keep working, it relies on player pool and .pool property is going to be obfuscated.

the reason there are so many cheaters is because you and the devs of this repository keep saying this lazy excuse of "oh we can't do anything about it because its a browser game and can't make anticheat on a browser game blah blah blah", who says it can't be done? who says it has to be an anticheat and not just a patch? it is very easy to detect surplus in its current state, you can detect it from the way it stores data, from the hook, from the properties it leaves on certain objects. or you could just patch it.

> The reason I brought up ASTs is to show that changing a few runtime behaviors is very unlikely to actually stop cheats If someone wants to get around this they don’t even strictly need AST tooling they could manually patch it but ASTs are the only reliable way to automate patching when code is minified, obfuscated, or reshuffled between builds

nobody is using AST for cheats (nor code patching), thats only old krity and I am pretty sure drino made new krity version either fully skidded from surplus's method or built the entire surplus client from scratch every time theres an update

please do more research before dropping paragraphs of spewed nonsense thats completely unrelated

> Regarding function.toString() integrity that detection only applies when patching happens at runtime (for example via userscripts or post-load hooks) In that case yes the browser has already parsed the code and toString() can reveal changes

that would be the case if there wasn't a hook on Function.prototype.toString, however yes it is possible to do that if you just make an iframe and use the iframe's Function.prototype.toString on the hooked function to get the real tostring

> As for the second point I agree that dismantling the current cheat ecosystem by breaking your code and by extension forks of it will have a real short term impact (granted though it will be like a week or so not 3 months and even if no one decides to fix it people will just move to the next cheat :o) my point is this "fix" can practically boil down to doing nothing due to how easy it is to bypass whether its a slightly capable dev or a script kiddie relying on a LLM

the next cheat in question being krity which is paid, nobody is going to pay to cheat on survev. you go once again on this pathetic mindset of not doing anything against cheats because theres just gonna be another one, well what if not what if its the last one or what if it will actually cause there to be less cheaters for once? which it very much likely will do that, I can assure you there will be cheaters after this but in far smaller numbers - making it easier both for players and moderators.

As for the next cheat that will appear, it will not be as competent as surplus was, surplus was far superior than what came before it. Krity got detected and crashed over and over because of simple dom detections or variables leaked in window object.

And that is because asset swapping and code patching are simply the most incompetent way to make a cheat, as it often forces you to leak variables in window and get detected.

Stopping the next cheat will be far easier than the current cheats.

leia-uwu:
popcorn

noam0001:
> popcorn

Hello miss brazil woman, leia! please review this request of a pulling at the nearest free space that may be given to you. Prosperity to brazil!

noam0001:
Long live the trees of the amazonas in brasilia

DamienVesper:
closing as stale

PR #344, opened 2026-01-25
refactor(wip): biome => oxlint+oxfmt;

lmssiehdev:
+ tsgo + rolldown


client/vite-plugins/codefendPlugin.ts
leia-uwu, line 1: somehow this and stripBlock.ts got formatted as 2 spaces instead of 4 lol

lmssiehdev: i didn't even lint the codebase yet, so probably my editor changed to the default config when i removed biome

PR #347, opened 2026-01-27
feat (modding): finish ModAPI v1.0 and docs

chickenpoo351:
This PR introduces the **initial V1.0 foundation** of the `survevModAPI`.

The focus of this V1.0 is intentionally conservative and limited

- `on*` **hooks** for observing and running code upon game / player events.
- `get*` **hooks** for querying game state (completely read only).
- **No** `set*` **hooks for now**

The absence of `set*` hooks is intentional.

The goal of this PR is meant to establish a stable, well documented, observation layer first, and gathering maintainer feedback before committing to any API hooks which modify game behavior or state

Now as for the motivation behind this.

Historically, mods for survev.io have relied on monkeypatching or full code replacement. While powerful, these approaches are brittle and prone to breaking across updates. This API aims to replace that with a stable interface, starting with safe primitives that wont lock the project into hard-to-reverse decisions.

An additional benefit of providing a well-defined modding sandbox is that it gives mod authors a legitimate and supported path for extending the game, reducing the need to learn invasive techniques such as monkeypatching or code replacement for common use cases.

Once this foundation is agreed upon, mutation hooks can be discussed and designed collaboratively between maintainers and mod authors.

To recap this PR includes

- Initial `on*` callback hooks
- Initial `get*` query hooks
- Complete documentation and examples of use
- A versioned changelog

Any feedback is very welcome and appreciated.

### DIFF
--- a/MOD_API_DOCS.md
+++ b/MOD_API_DOCS.md
@@ -1,0 +1,91 @@
+# Introduction
+
+Hi! This is a ModAPI created for survev.io, designed to provide a stable and predictable way for mods to interact with the game.
+
+Mods interact with the game through event hooks (on*), read-only accessors (get*), and a small set of intentionally limited modification hooks (set*).
+
+# How Mods Work
+
+Mods register callbacks using on* hooks. These hooks fire in response to specific game events, allowing mods to react by running their own logic.
+
+(Yes, you *can* try to make a dancing chicken appear each time someone dies. No judgement here...)
+
+Mods can read certain pieces of game state using get* functions. All data returned by these functions is read-only and represents a snapshot of the current state.
+
+(Think of it like looking through your neighbors window, nice view but you can't take their TV.)
+
+In limited cases, mods may also modify certain aspects of the game using set* functions. These APIs are intentionally constrained to prevent instability or unintended side effects.
+
+(Alas you cannot replace textures that other players (without your mod) would see. So your dream of making the whole map completely filled with water will have to wait...)
+
+# History / Background
+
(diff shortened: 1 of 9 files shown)


chickenpoo351:
> There is a lof of discussion about the implementation details but none about the usecase, do u run a fork or maintain some userscript that would take advantage of this? What are some examples?

I don’t currently maintain a public fork, but I have written a few client-side mods / custom clients for the game. In nearly every case, they ran into the same issue:

- They relied on monkeypatching internal functions

- Or replacing chunks of game code directly

- And they broke across updates

Right now, if someone wants to build tooling or experimental features on top of the client, the only realistic path is invasive patching. That’s brittle, hard to maintain, and pushes people toward techniques that are structurally similar to exploit development.

My thought process with this PR is:

If we provide a stable, read-only observation layer, we give people a legitimate extension path that:

- Doesn’t grant extra authority

- Doesn’t expose new server-side trust assumptions

- Doesn’t require patching internals

Is easier for maintainers to reason about long term

This won’t stop cheat creators that’s unrealistic but it does lower the barrier for people who just want to build creative or analytical tooling without reverse engineering the client or learning invasive techniques specifically for this game.

Some concrete use cases this API would already enable:

- UI overlays (or at least more advanced ones than currently capable)

- Custom match analytics / stat tracking

- Accessibility features

- General experimentation without reverse engineering the client

If it would help ground the discussion, I can put together a small example mod built on top of these hooks to demonstrate a real use case. Of course though I would need a few days to do so as I am currently stacked with a few projects

chickenpoo351:
Hi @leia-uwu I wanted to see how the review process on this PR has been going... it has been nearly half a year since your last interaction with the PR and I just want to know if you actually want to accept this or if you only like parts of it or just don't like it at all instead of having the PR kinda sit in this limbo state

DamienVesper:
Maybe will revisit this in the future. As of right now, not planning on implementing.

PR #349, opened 2026-02-08
refactor: set Team Ping Menu to default keybind

InkyDarkBird:
Why does Team Ping Hold still exist? It is the only keybind in survev.io that requires 2 separate inputs to be used **at the same time AND in a specific order** to activate, which is never explained to you in the game whatsoever. Furthermore, Team Ping Hold never makes any indication that it is somehow related to the Emote Menu keybind. Because Team Ping Menu is objectively easier to use than Team Ping Hold, I fail to see why anyone who would understand the difference between Team Ping Hold and Team Ping Menu would ever want to use Team Ping Hold besides the fact that they have been using Team Ping Hold for a long time.

This is extremely unintuitive for new players, as I would assume that the majority of people assume that, unless clearly specified otherwise, that a keybind only requires 1 input in order to activate.

Team Ping Hold should be removed. If it isn't removed, then Team Ping Menu should at least replace it as the default keybind for C.

### DIFF
--- a/shared/defs/maps/baseDefs.ts
+++ b/shared/defs/maps/baseDefs.ts
@@ -701,6 +701,7 @@ export const Main: MapDef = {
             { name: "martyrdom", count: 1, weight: 1 },
             { name: "self_revive", count: 1, weight: 1 },
             { name: "bonus_9mm", count: 1, weight: 1 },
+            { name: "high_velocity", count: 1, weight: 1 },
         ],
         tier_potato_perks: [
             { name: "", count: 1, weight: 25 },


shared/gameConfig.ts
leia-uwu, line 1: we cant remove or reorder enum values in that enum without breaking all existing saved keybinds on user's configs, as the keybinds are serialized using the values of the enums

DamienVesper: well you can...it just won't be a pretty migration. nevertheless i agree with the default change rather than remove alltogether

PR #359, opened 2026-03-09
feat: pass

lmssiehdev:
GET THIS MERGED BEFORE SURVIV

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -37,7 +37,6 @@ import { ObjectType } from "../../../../shared/net/objectSerializeFns";
 import type { GroupStatus } from "../../../../shared/net/updateMsg";
 import { type Circle, coldet } from "../../../../shared/utils/coldet";
 import { collider } from "../../../../shared/utils/collider";
-import type { Loadout } from "../../../../shared/utils/loadout";
 import { math } from "../../../../shared/utils/math";
 import { assert, util } from "../../../../shared/utils/util";
 import { type Vec2, v2 } from "../../../../shared/utils/v2";
@@ -226,13 +225,16 @@ export class PlayerBarn {
             ip,
             joinData.findGameIp,
             joinData.userId,
-            joinData.loadout,
             joinData.quests,
         );
 
         this.socketIdToPlayer.set(socketId, player);
 
         this.activatePlayer(player, group, team);
+        player.setLoadout(
+            joinData.loadout ? joinData.loadout : joinMsg.loadout,
+            !joinData.loadout,
+        );
 
         return player;
     }
@@ -824,8 +826,14 @@ export class Player extends BaseGameObject {
 
     setOutfit(outfit: string) {
         if (this.outfit === outfit) return;
-        this.outfit = outfit;
         const def = GameObjectDefs[outfit] as OutfitDef;
+        if (this.game.map.factionMode) {
+            if (def.teamId && this.teamId !== def.teamId) {
+                return;
+            }
+        }
+        this.outfit = outfit;
+
         this.obstacleOutfit?.destroy();
         this.obstacleOutfit = undefined;
 
@@ -1409,7 +1417,6 @@ export class Player extends BaseGameObject {
         ip: string,
         findGameIp: string,
         userId: string | null,
-        loadout?: Loadout,
         questIds?: string[],
     ) {
         super(game, pos);
@@ -1490,8 +1497,6 @@ export class Player extends BaseGameObject {
         for (const [item, amount] of Object.entries(defaultItems.inventory)) {
             this.invManager.set(item as InventoryItem, amount);
         }
-
-        this.setLoadout(loadout ? loadout : joinMsg.loadout, !loadout);
 
         if (this.game.map.sniperMode) {
             this.invManager.give("2xscope", 1);
@@ -4103,6 +4108,11 @@ export class Player extends BaseGameObject {
                 }
                 break;
             case "outfit":
+                if (this.game.map.factionMode) {
+                    if (def.teamId && this.teamId !== def.teamId) {
+                        return;
+                    }
+                }
                 if (this.role) {
                     const roleDef = GameObjectDefs[this.role] as RoleDef;
                     if (roleDef.defaultItems?.noDropOutfit) {
--- a/shared/defs/gameObjects/outfitDefs.ts
+++ b/shared/defs/gameObjects/outfitDefs.ts
@@ -32,6 +32,9 @@ export interface OutfitDef {
     obstacleType?: string;
     baseScale?: number;
     ghillie?: boolean;
+    // if on faction mode, the skin will only work for the specified ID
+    // 1 for red team, 2 for blue team
+    teamId?: number;
 }
 
 function defineOutfitSkin(baseType: string, params: DeepPartial<OutfitDef>): OutfitDef {
@@ -91,6 +94,7 @@ const SkinDefs: Record<string, OutfitDef> = {
             sprite: "loot-shirt-01.img",
             tint: 0xffffff,
         },
+        teamId: 1,
     }),
     outfitTank: defineOutfitSkin("outfitBase", {
         noDrop: true,
@@ -159,6 +163,7 @@ const SkinDefs: Record<string, OutfitDef> = {
             sprite: "loot-shirt-01.img",
             tint: 0xffffff,
         },
+        teamId: 2,
     }),
     outfitAssault: defineOutfitSkin("outfitBase", {
         noDrop: true,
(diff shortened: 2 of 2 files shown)


The-Overseer1:
hahahahahahahahahaha

server/src/utils/playerData.ts
```diff
@@ -0,0 +1,43 @@
...
+                          .from(usersTable)
+                          .leftJoin(
+                              userQuestTable,
+                              and(eq(userQuestTable.userId, usersTable.id)),
+                          )
+                          .where(inArray(usersTable.id, userIds))
+                          .groupBy(usersTable.id)
+                  ).map((r) => [r.userId, r]),
+              )
+            : {};
```
leia-uwu: can you extract this to an if instead of a ternary and the db call to a variable, to reduce the nesting

lmssiehdev:
this has been running on resurviv for a while now, @NAMERIO reported no issues besides the quest not refetching consistently when players leave

the client only updates when you quit through the game end button, it's probably better to move the update pass logic inside quitGame so it's refetched universally every time the game ends, but it's beyond the scope of this PR

https://github.com/survev/survev/blob/31d6e1dbfe65106f832684398804e69e1da84fb7/client/src/ui/ui.ts#L374-L384

another thing that's worth noting is that the client hardcodes pass_survivr1 and so does the server, this would need some minor refactoring to support multiple passes, I might fix it if I have the time but it shouldn't be a blocker for this getting merged, especially since the game's future is uncertain now

server/src/utils/playerData.ts
lmssiehdev: @GROK

lmssiehdev:
```
>create an account two hours ago*
>sends me a notification
>refuses to elaborate
>leaves
```

server/src/api/routes/user/PassRouter.ts
```diff
@@ -0,0 +1,291 @@
...
+});
+
+PassRouter.post(
+    "/set_quest",
+    validateParams(zSetQuestRequest),
+    // biome-ignore lint/suspicious/useAwait: to please the client
+    async (c) => {
+        return c.json({ success: true }, 200);
+    },
+);
```
leia-uwu, line 274: is this actually used? the client `setQuest` doesn't seem to be called anywhere

leia-uwu:
> the client only updates when you quit through the game end button, it's probably better to move the update pass logic inside quitGame so it's refetched universally every time the game ends, but it's beyond the scope of this PR

actually turns out we need to send and `UpdatePassMsg` to the client so it refreshes the quest once quitGame is ran

server/src/api/routes/user/PassRouter.ts
leia-uwu: seems like it was maybe used for debugging on dev builds?

so you could set a quest type from the client to easily test it

leia-uwu:
@lmssiehdev mind reviewing my commits when you have time? to make sure i didnt do anything dumb with drizzle and the bot

client/src/ui/pass.ts
```diff
@@ -125,116 +126,117 @@ export class Pass {
...
-                // Humanize time for timed quests
-                let targetText: string | number = quest.data.target;
-                if (questDef.timed) {
-                    targetText = humanizeTime(targetText);
-                }
+            // Humanize time for survival quests
+            let targetText: string | number = quest.data.target;
+            if (questDef.event === "survived") {
+                targetText = humanizeTime(targetText);
+            }
```
leia-uwu, line 215: I'd prefer to keep the `questDef.timed` property over hardcoding it to only survived time quests

server/src/api/routes/user/PassRouter.ts
lmssiehdev: > seems like it was maybe used for debugging on dev builds?

that's what i thought as well

bot/src/commands/helpers.ts
```diff
@@ -20,7 +21,8 @@ export function createCommand<T extends z.ZodSchema = z.ZodSchema>(config: {
             | ApplicationCommandOptionType.String
             | ApplicationCommandOptionType.Integer
             | ApplicationCommandOptionType.Boolean
-            | ApplicationCommandOptionType.User;
+            | ApplicationCommandOptionType.User
+            | ApplicationCommandOptionType.Integer;
```
lmssiehdev: duplicated `ApplicationCommandOptionType.Integer`

bot/src/commands/helpers.ts
```diff
@@ -112,6 +115,9 @@ export function createSlashCommand(config: ReturnType<typeof createCommand>) {
             case ApplicationCommandOptionType.User:
                 builder.addUserOption((opt) => configureBuilderOption(opt, option));
                 break;
+            case ApplicationCommandOptionType.Integer:
+                builder.addIntegerOption((opt) => configureBuilderOption(opt, option));
+                break;
```
lmssiehdev: same here

server/src/api/routes/private/passXp.ts
```diff
@@ -0,0 +1,79 @@
+import { and, eq, sql } from "drizzle-orm";
+import { GameObjectDefs } from "../../../../../shared/defs/gameObjectDefs";
+import { PassDefs } from "../../../../../shared/defs/gameObjects/passDefs";
+import { passUtil } from "../../../../../shared/utils/passUtil";
+import { Config } from "../../../config";
+import { server } from "../../apiServer";
+import type { db } from "../../db";
+import { itemsTable, userPassTable } from "../../db/schema";
+
+export async function incrementPassXp(
```
lmssiehdev, line 10: it's better to extract the whole /quest_progress route instead of just the helper function, so it's consistent with the other files inside the /routes/ folder being actual routes

bot/src/commands/helpers.ts
```diff
@@ -30,10 +32,16 @@ export async function genericExecute<N extends Exclude<Command, "search_player">
     name: N,
     interaction: ChatInputCommandInteraction,
     validator: z.ZodTypeAny,
+    requiresAdmin = false,
     isPrivateRoute = false,
```
lmssiehdev, line 35: requiresAdmin and isPrivateRoute do the same thing no? why have both?

bot/src/commands/helpers.ts
```diff
@@ -74,11 +82,6 @@ async function handlePrivateRoute(
     name: any,
```
lmssiehdev: `handlePrivateRoute` is probably not needed anymore, just replace it with `const client = requiresAdmin ? honoClient[name] : honoClient.moderation[name];` we are ignoring the types so typescript won't complain

bot/src/commands/helpers.ts
leia-uwu: ~~i really need to stop coding while being sleep deprived~~

bot/src/commands/helpers.ts
leia-uwu: no
`isPrivateRoute` now just makes it use `/private` instead of `/private/moderation` for the command
`requiresAdmin` does the discord permission check

i separated both because i added 2 routes to `/private/moderation` that require admin

server/src/api/routes/private/passXp.ts
leia-uwu: i should just move the file to outside routes i suppose, i extracted the function because theres more than just quests that can give XP (example: the moderation command and xp items)

don't really see a point on extracting the route

leia-uwu: ~~also not all files are routes there already, server/src/api/routes/user/auth/authUtils.ts is not~~

bot/src/commands/helpers.ts
lmssiehdev: i see

server/src/api/routes/private/passXp.ts
lmssiehdev: fair, they should probably be moved out but it's not a deal breaker

client/src/ui/pass.ts
lmssiehdev: can you change this? not sure if i'll get to this anytime soon

unrelated by I got recording of multiple POVs to work after with a custom ReplayMsg that records everything, it's currently not optimized or anything, 8mb per game lol, I can make a PR if you want to take a look otherwise it will just get lost next time i try and clean my system to save space

leia-uwu: > unrelated by I got recording of multiple POVs to work after with a custom ReplayMsg that records everything, it's currently not optimized or anything, 8mb per game lol, I can make a PR if you want to take a look otherwise it will just get lost next time i try and clean my system to save space

mamoun is working on the exact same thing so he might be interested lol

PR #381, opened 2026-03-30
perf: optimize object pool allocation and iteration

jkoppel:
## Summary
This PR optimizes the core `Pool` class to provide O(1) allocation and O(N_active) iteration. Previously, the engine used a linear scan for allocation and iterated over all historical objects (including inactive ones) in most update/collision loops, leading to performance degradation in long game sessions.

## Changes
- **O(1) Allocation**: Added `m_freeList` stack to `Pool` to avoid O(N) linear scans when allocating new objects.
- **O(1) Removal**: Implemented swap-and-pop strategy using a new `__poolIdx` field on `AbstractObject`. This allows removing objects from the active pool in constant time without `splice`.
- **O(N_active) Iteration**: Modified `m_getPool()` to return an array containing only active objects. This automatically optimizes over 40 call sites (physics, rendering, updates) by eliminating the need to check the `active` flag on thousands of dead entities.
- **Cleaned up loops**: Removed redundant `if (obj.active)` checks in hot paths where `m_getPool()` is now guaranteed to return active objects.

## Performance Impact
Simulated a 30-second squad fight (4 players, high rate-of-fire weapons) in a session with 100,000 historical objects processed:
- **Old Implementation**: ~2.09ms overhead per update cycle
- **New Implementation**: ~1.82ms overhead per update cycle
- **Spawning Cost**: Reduced from O(N_total) to O(1).

Benchmark results: [GitHub Gist](https://gist.github.com/jkoppel/39e541099314db9c0e922a37e5f3df7e)


IOMAN1:
is this vibe coded?

leia-uwu:
also the commit author is "undefined", which is really suspicious..., when i first saw this PR my reaction was "is this an openclaw agent????"

ignoring that, i'm not sure this is even worth it? 2.09ms to 1.82ms with 100k objects doesn't seem like it would matter in any realistic scenario

the mode with the most amount of objects (woods) doesn't even come close to 10k objects during map generation, and no client is ever going to see all the objects of the entire map at once, 100k is just not a realistic scenario lol

not to mention that the max amount of objects is around 65k because object IDs are an uint16 lol

jkoppel:
Hi all! My thinking was "I want to contribute to this project. I've found no way to talk to the devs other than here, andI have no idea what the maintainers are prioritizing right now, no idea how they'd respond to more tests, refactors, features, etc....but everyone likes speed, so I'm going to find some really uncontroversial but real improvement of under 100 lines." And this is what I found.

But now I'm here talking to the devs, and am eager for my talents to be directed to more pressing concerns.


The benchmark is 100% vibe-coded. The actual code change is AI-assisted but very much not vibe coded, having gone through high levels of review and editing.

ei-pi:
what
at first I thought this was like, a tutorial video thingy for how to play the game that you added for new players and was really confused why it was on this pr
but it's actually a 6-second clip that I'd imagine in the marketing for an ai editor
what exactly does this show or add?

the clip, for anyone who doesn't feel like downloading ts

https://github.com/user-attachments/assets/7a3367ba-56e7-49f4-9f3b-a34d719a0206

jkoppel:
Ultra embarassing. Big oops.

That was meant to be committed to a different repo. Removing.

PR #386, opened 2026-04-05
enhance: replace most Arial instances with Roboto Condensed, slight tweak to scope icons

SquaredHexahedron:
Made the tick lengths on the scope icons consistent

Didn't change [client/css/dev.css](https://github.com/survev/survev/blob/master/client/css/dev.css) since I think dev-only stuff not being in Roboto Condensed isn't that big of a deal


client/public/privacy.html
```diff
@@ -7,7 +7,7 @@
   <title>Privacy Policy - Survev.io</title>
   <style>
     body {
-      font-family: Arial, sans-serif;
```
leia-uwu, line 10: on this page it will only work if you have the font installed system wide, you need to make it load the font like the main index.html does

SquaredHexahedron: Interesting; I'm actually kinda fine with leaving that use Arial since it's a privacy policy, on a different page, and pretty distinct from the main game

SquaredHexahedron: But sure why not I made it load the font

PR #394, opened 2026-04-12
refactor: rename armorPiercing and armorPlated properties to  canBreakReinforced and isReinforced

InkyDarkBird:
Now that AP Rounds exist, I believe that it is extremely misleading for the armorPiercing property in melee weapons to refer to breaking ammo crates. I refactored it to be cratePiercing to better align with how the property pierces ammo crates, not armor.

### DIFF
--- a/server/src/game/objects/obstacle.ts
+++ b/server/src/game/objects/obstacle.ts
@@ -366,21 +366,21 @@ export class Obstacle extends BaseGameObject {
         if (this.health === 0 || !this.destructible) return;
 
         if (params.damageType === DamageType.Player) {
-            let armorPiercing = false;
+            let canBreakReinforced = false;
             let stonePiercing = false;
 
             if (params.gameSourceType) {
                 const sourceDef = GameObjectDefs[params.gameSourceType] as
                     | {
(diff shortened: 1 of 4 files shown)


DamienVesper:
this is not me expressing distaste for this PR, but i could not help but think of kaklik when i saw this

singe725:
To be fair, a few of the naming conventions are a bit confusing

leia-uwu:
`cratePiercing` is an even more misleading name imo

99% of crates in the game do not need a `cratePiercing` melee to break, only `armorPlated` obstacles do, and those dont need to be "crates", and also the `armorPlated` property on obstacles was not renamed in this MR, making the misleading situation even worse because now you have `stonePlated` that requires `stonePiercing` melees and `armorPlated` obstacles that require `cratePiercing` to break, making the names mismatch and be inconsistent

leia-uwu:
fwiw i'm not against renaming it, but `cratePiercing` is just not it and makes the situation worse

InkyDarkBird:
buh
I changed it to `reinforced` and `canBreakReinforced`

PR #400, opened 2026-04-22
Add additional patterns to the username filter

Abubu-ok:
from my understanding, this is mostly enough to filter Epstein out of the game, tell me if it isn't like that.
hopefully, this will reduce the number of ban game mods have to make.


SquaredHexahedron:
I wonder if replacing some of the patterns with regexes would help

Abubu-ok:
you mean the part I edit or the original patterns or both? I thought I just had to add a few simple lines then RegExpMatcher will handle everything else, is that not the case? tbh, I don't really understand much of the code.

lmssiehdev:
> I wonder if replacing some of the patterns with regexes would help


https://github.com/jo3-l/obscenity/blob/main/docs/reference/functions/pattern.md

PR #402, opened 2026-05-03
Use uWebSockets.js collectBody helper function

evanwild:
uWebSockets.js added a collectBody helper for reading all chunks of a request into a single buffer:

> collectBody is a helper function making optimal use of the new onDataV2. It allows efficient and easy collection of smallish HTTP request body data into RAM. It accumulates all data chunks and calls handler with the complete body as an ArrayBuffer once all data has arrived. If the total body size exceeds maxSize bytes, handler is called with null instead.

This PR removes the custom readPostedJSON function and uses collectBody instead. It has the added benefit of having a max size to prevent someone allocating a lot of memory.


leia-uwu:
thanks! i have been wanting to get rid of that horrible readPostedJSON method for a while lol

PR #403, opened 2026-05-03
Fix/loot spawning

leia-uwu:
@InkyDarkBird i cant add you as a reviewer so i will just ping you instead

deploy preview has a bazilion crates and airdrops on normal mode for testing purposes

### DIFF
--- a/server/src/game/objects/obstacle.ts
+++ b/server/src/game/objects/obstacle.ts
@@ -457,13 +457,13 @@ export class Obstacle extends BaseGameObject {
             v2.set(lootPos, v2.add(this.pos, v2.rotate(def.lootSpawn.offset, this.rot)));
         }
 
-        const loot = [...def.loot];
+        const lootTablesOrItems = [...def.loot];
 
         if (
             params.source?.__type === ObjectType.Player &&
             params.source.hasPerk("scavenger")
         ) {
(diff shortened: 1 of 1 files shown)


leia-uwu:
(test is failing because of the debug commit lol)

InkyDarkBird:
I think the majority of the interactions are correct, but I can't really test airdrops since 50v50 isn't in rotation on surviv.io.

I think one small difference is that the player punch exerts a slightly larger velocity on the loot drops in surviv.io than survev.io, at least for Cobalt class pods.

The following videos below show a examples of this.

https://github.com/user-attachments/assets/2be189ef-7f25-49ef-9e64-ec734c34cb61

https://github.com/user-attachments/assets/384fd3b7-2f3a-49bf-8299-c329fb62c3eb

I don't think this affected Soviet crates, although that could be because the guns spawn loot alongside them and dissipate the velocity more evenly.

InkyDarkBird:
I mean unless airdrops are super inaccurate I don't think it matters tooo much since it feels smooth enough

leia-uwu:
merging this as its an improvement over the current one on master, further improvements would require ~~ioman's purpose in life to be fulfilled~~ fully fixing loot physics

PR #404, opened 2026-05-06
feat: replace publift ads with nitropay

lmssiehdev:


lmssiehdev:
image


i deployed an actual working versrion with test ads, feel free to test, all ads are visible for testing but they can be disabled when needed, check the loadout as well

https://survev-ads-test.vercel.app/stats/?slug=preacher?nitro_demo=1
https://survev-ads-test.vercel.app/stats?nitro_demo=1
https://survev-ads-test.vercel.app/?nitro_demo=1

leia-uwu:
image

if it fails to load it seems to crash the entire client and prevent it from loading

leia-uwu:
ok thats funny:
if i have ublock origin enabled, it will block the ads in a way that doesn't crash the client

if i disable ublock origin, firefox tracking protection will block the ads but in a way that crashes the client

leia-uwu:
also if we are adding ads to the /stats app can we remove at least one from the main menu, like the one on top of the logo, it also doesn't resize very well:

image

lmssiehdev:
>  firefox tracking protection will block the ads but in a way that crashes the client

nice catch, please recheck again if it's fixed

>  like the one on top of the logo, it also doesn't resize very well  

this is only an issue on mobile sized viewports and the top ad gets hidden when the game detects mobile, still i added a media query to hide it on smaller heights just to be safe

if the stats ads are an issue feel free to disable them, i just added them so the implementation is finished instead of the commented out dead code

also, i fixed an unrelated issue
image

PR #408, opened 2026-05-11
Refactors and cleanups step 1

leia-uwu:
- [x] Step 1: add import extensions

- [x] Step 2: switch away from biome to dprint + oxlint (oxfmt sucks so we are not using it, its just prettier but in rust and prettier sucks)

- [x] Step 3: Remove direct access to GameObjectDefs and MapObjectDefs in favor of a class with 2 methods:
- One that returns the definition type or undefined, and one that automatically asserts if the definition exists, will make accessing definitions more safe and give better errors


lmssiehdev:
GitHub loves dismissing my reviews unless it's a 280 changed files PR

PR #409, opened 2026-05-13
fix: make cobalt class pods not change owners

InkyDarkBird:
https://discord.com/channels/1407084649343352852/1503860436502581289

This fix prevents other players from being falsely labeled as the last person who interacted with the class pod if they try to interact with the class pod after the original person opened it.

### DIFF
--- a/server/src/game/objects/obstacle.ts
+++ b/server/src/game/objects/obstacle.ts
@@ -654,8 +654,6 @@ export class Obstacle extends BaseGameObject {
             return;
         }
 
-        this.interactedBy = player;
-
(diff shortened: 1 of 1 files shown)


server/src/game/objects/obstacle.ts
```diff
@@ -677,6 +677,7 @@ export class Obstacle extends BaseGameObject {
 
         if (this.isButton && this.button.canUse) {
             this.useButton(player);
+            this.interactedBy = player;
```
leia-uwu, line 677: should go before the `useButton` call

PR #410, opened 2026-05-13
Fix some quest related bugs

leia-uwu:
game server was not reporting errors correctly and there was a race condition that could end up with the user getting duplicated quests:
image

### DIFF
--- a/server/src/api/routes/user/PassRouter.ts
+++ b/server/src/api/routes/user/PassRouter.ts
@@ -136,11 +136,12 @@ async function rerollSlot(
     now: number,
     rerolled: boolean,
     loadedQuests: UserQuestTableSelect[],
+    transaction?: Parameters<Parameters<typeof db.transaction>[0]>[0],
 ) {
     const excludedTypes = new Set(loadedQuests.map((quest) => quest.questType));
     const questType = getRandomQuestType(excludedTypes);
 
-    await db
+    await (transaction ?? db)
         .update(userQuestTable)
         .set({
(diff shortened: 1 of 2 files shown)


lmssiehdev:
nice catch, the solution is more complicated than it should be tho, it can probably be solved with a `FOR UPDATE` lock on the db

(we use it to solve similar issue with a quest in a codebase i'm working on)
image

server/src/api/routes/user/PassRouter.ts
```diff
@@ -236,32 +237,54 @@ PassRouter.post("/get_pass", validateParams(zGetPassRequest), async (c) => {
...
+        const now = Date.now();
+        const expired = quest.nextRefreshAt - now < 0;
+        const refreshEnabled = (!quest.rerolled && !quest.complete) || expired;
+        if (!refreshEnabled) {
+            return false;
+        }
+        await rerollSlot(user.id, idx, now, !expired, quests, transaction);
+        return true;
+    });
+    console.log(idx, "end");
```
lmssiehdev: :p

lmssiehdev: beside this it looks good to my tired eyes

leia-uwu: whops lol, i'm also tired :p

lmssiehdev (dismissed):

leia-uwu:
kinda annoying how touching the branch just dismisses the review lol

PR #411, opened 2026-05-14
Update ko.json

Outliers1211:
korean translation for the new update (perks)


The-Overseer1:
@Outliers1211 see #412 for further changes to perk descriptions (Hyperfrag, Combat Stims, Flak Jacket) as well as previous changes to en.json (cobalt, Event Winnr, etc)

but otherwise ty :D

PR #418, opened 2026-05-17
fix: make throwable keybind not automatically launch throwable if pla…

InkyDarkBird:
…yer has no other throwables

https://github.com/user-attachments/assets/87e72171-06de-499c-b909-5cc576b04074

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -47,7 +47,7 @@ import type { Game, JoinTokenData } from "../game";
 import { Group, Team } from "../group";
 import { InventoryManager } from "../inventoryManager";
 import { QuestManager } from "../questManager";
-import { WeaponManager } from "../weaponManager";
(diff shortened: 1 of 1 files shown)


server/src/game/objects/player.ts
```diff
@@ -3738,12 +3738,21 @@ export class Player extends BaseGameObject {
...
+                                this.invManager.has(t as InventoryItem));
+
+                        if (hasOtherThrowables) {
+                            this.weaponManager.throwThrowable(true);
+                            if (this.animType === GameConfig.Anim.Cook) {
+                                this.cancelAnim();
+                            }
+                            this.weaponManager.showNextThrowable();
+                        } else {
+                            break;
```
leia-uwu, line 3754: hmm not sure if i like this here
feels like we should maybe move the throwing code to inside `showNextThrowable` and just let that function handle it

PR #421, opened 2026-05-21
fix: hide ads instead of unmounting them

lmssiehdev:


lmssiehdev:
the prev implementation is not generating any money so this is urgent, can you merge if this looks good?


i'm deploying a testing url in the meantime (edit: couldn't but the changes are straightforward)

leia-uwu:
im just going to trust you that it works ig since i dont have enough context with how the ads work

PR #424, opened 2026-05-22
fix: remove pod ownership upon death or disconnect

InkyDarkBird:
https://discord.com/channels/1407084649343352852/1506763998270591107

If the pod owner dies before they can pick up the loot, the game still treats them as if they exist and will prevent other players from picking up the pod loot temporarily. This commit removes that post-mortem ownership.

### DIFF
--- a/server/src/game/objects/loot.ts
+++ b/server/src/game/objects/loot.ts
@@ -340,8 +340,13 @@ export class Loot extends BaseGameObject {
 
     update(dt: number): void {
         if (this.hasOwner) {
+            const owner = this.game.objectRegister.getById(this.ownerId);
             this.removeOwnerTicker += dt;
-            if (this.removeOwnerTicker > 2) {
+            if (
+                this.removeOwnerTicker > 2 ||
+                !owner ||
+                (owner.__type === ObjectType.Player && (owner.dead || owner.disconnected))
+            ) {
                 this.ownerId = 0;
                 this.setDirty();
             }


server/src/game/objects/loot.ts
```diff
@@ -340,8 +340,13 @@ export class Loot extends BaseGameObject {
 
     update(dt: number): void {
         if (this.hasOwner) {
+            const podUnlocker = this.game.objectRegister.getById(this.ownerId);
```
leia-uwu: just `owner` would be a better variable name, i know its currently only used for pods, but if in the future its not then it becomes a misleading name

server/src/game/objects/loot.ts
```diff
@@ -340,8 +340,13 @@ export class Loot extends BaseGameObject {
...
             this.removeOwnerTicker += dt;
-            if (this.removeOwnerTicker > 2) {
+            if (this.removeOwnerTicker > 2 ||
+                (podUnlocker &&
```
leia-uwu: also i feel like if the owner doesn't exist anymore the owner ID should also be removed

server/src/game/objects/loot.ts
```diff
@@ -340,8 +340,13 @@ export class Loot extends BaseGameObject {
 
     update(dt: number): void {
         if (this.hasOwner) {
+            const owner = this.game.objectRegister.getById(this.ownerId);
             this.removeOwnerTicker += dt;
-            if (this.removeOwnerTicker > 2) {
+            if (
+                this.removeOwnerTicker > 2 ||
+                !owner ||
+                (owner && owner.__type === ObjectType.Player && owner.dead)
```
leia-uwu:
```suggestion
                (owner.__type === ObjectType.Player && (owner.dead || owner.disconnected))
```

PR #434, opened 2026-06-03
feat: allow lone survevrs to ping each other

InkyDarkBird:


leia-uwu:
perhaps it should be a property in the role definition in the future but for now its fine...

PR #437, opened 2026-06-04
Misc server cleanups

leia-uwu:
decided to do this before working more on #422 to make my life easier

no more single threaded game server (having 2 different code paths was annoying and would often cause bugs only present in one of the code paths)
no more server plugins (was basically unmaintained, half baked and broken)

less stuff to maintain and to worry about :)


lmssiehdev:
> no more server plugins  

rip resurviv.biz

leia-uwu:
> > no more server plugins
> 
> rip resurviv.biz

do they actually use the unmodified version of the plugin manager from the main repo? i'm not sure that thing even works because it has been completely unmaintained with nothing that tests it, and it was so incomplete and badly made that i decided that people would have a harder time trying to make anything useful with it compared to just forking the game

lmssiehdev:
> do they actually use the unmodified version of the plugin manager from the main repo? 

i was using it up until the default export refactor which broke it for me so i just modified it a bit to get around that, i haven't touched resurviv in ages tho so not sure what they are using rn.

PR #440, opened 2026-06-06
fix: set noPotatoSwap to true for tomato

InkyDarkBird:


leia-uwu:
that doesn't solve the problem...
you should still be able to swap to tomatoes on PvT just not on regular potato mode, this just inverts the problem :/

DamienVesper:
closing this for just being wrong tbh, reopen it when you inevitably fix it

PR #441, opened 2026-06-07
Run each game process on its own port

leia-uwu:
Should help with server lag and stutters a lot in my testing

with 20 games with around ~900 stress test bots it was completely smooth on my system, with the old main process websocket proxy it was really laggy and would stutter a lot with that amount of players, with constant lag spikes because something blocking the main server process would create a lag spike across all games for all players

Note that this is a breaking change for server hosters since they now have to open a range of ports for games (defaults to port 9000 to 9063, max of 64 games by default)

Not sure if i should still keep support for the main port proxy, its not hard to support both so i kept removing it on a separate commit

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -21,6 +21,8 @@ export function getConfig(isProduction: boolean, dir: string) {
             port: 8001,
             apiServerUrl: "",
             thisRegion: "local",
+            firstGamePort: 9000,
+            maxGames: 64,
         },
         vite: {
             host: "127.0.0.1",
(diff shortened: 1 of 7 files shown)


server/src/utils/types.ts
```diff
@@ -1,8 +1,7 @@
 import { z } from "zod";
 import type { MapDefs } from "../../../shared/defs/mapDefs.ts";
 import { TeamMode } from "../../../shared/gameConfig.ts";
-import type { FindGameError } from "../../../shared/types/api.ts";
-import { loadoutSchema } from "../../../shared/utils/loadout.ts";
+import { type FindGameError, loadoutSchema } from "../../../shared/types/api.ts";
```
lmssiehdev: W

PR #449, opened 2026-06-21
Refactor/remove proxycheck ts

leia-uwu:


server/src/utils/fetchWithRetry.ts
```diff
@@ -0,0 +1,27 @@
+import { defaultLogger } from "./logger.ts";
+
+export async function fetchWithRetry(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    const maxAttempts = 3;
```
lmssiehdev, line 4: at this point the codecase can make use of something like better-result to help get rid of smelly parts like this

https://better-result.dev/advanced/retry-logic#basic-retry-configuration

leia-uwu: i'm not adding a new dependency for something thats like 15 lines of code

DamienVesper: It's preferable to do this in-house rather than add a dependency for a one-off use. Personally, I think current implementation is fine.

lmssiehdev: > one-off use  

> 15 lines of code

ig my comment wasn't clear enough but i wasn't referring to just this specific chunk, it will help clean a lot of ugly try catches across the backend with https://better-result.dev/core/pattern-matching for example

and it's possible to do it in house too

PR #451, opened 2026-06-21
Refactor/yeet esbuild

leia-uwu:
Esbuild generates kinda... stupid... bundles, never removes unused variables like unused zod objects from the bundle, and for some reason imports packages multiple times...?:
image

Anyway, vite now uses rolldown by default, reason i originally used esbuild was "because it was the same thing vite used so it was technically not a new dependency"

while the later is still true because vitest still uses esbuild for now, adding rolldown is also not a new dependency :)

but anyway rolldown produces smaller bundles, gameProcess.js went from 1.6mb to 1.3mb!!


leia-uwu:
🙃 nvm turns out nodejs completely removed `--experimental-transform-types` in nodejs 26, with dumb reasons for the removal...

i hate nodejs...

it still has `--experimental-strip-types` (which is not experimental anymore and is enabled by default) but it requires `erasableSyntaxOnly` on typescript, which means no enums, no namespaces and no `public`/`readonly`/`private` constructor arguments...

lmssiehdev:
> 🙃 nvm turns out nodejs completely removed --experimental-transform-types in nodejs 26

bun would never..

PR #459, opened 2026-06-28
fix: force switch to melee slot when downed

vyx77:
mentioned in: https://discord.com/channels/1407084649343352852/1520707991823847535
also fixes the related bug where the burst bullets don't get removed from their queue after being downed, so when revived they will continue shooting out the rest of the burst


leia-uwu:
thanks!

PR #472, opened 2026-07-16
feat/refactor: add 91 new flag emotes, reorganize emotes in files, fix some existing emotes, update Japanese l10ns

SquaredHexahedron:


ei-pi:
a) where did you get all these
b) is that every country lol (except The One That Shall Not Be Named™)

leia-uwu:
Honestly adding that many flags will make browsing the emote selector be really frustrating and hard to find specific flags

Don't think this can be merged until the loadout menu has a search bar functionality...

leia-uwu:
since the reorganizing is on its own commit and before the new flags, you could put that in a different merge request (or remove the new flags from this one and move those to a different MR) so that commit can land first

SquaredHexahedron:
> a) where did you get all these 
> b) is that every country lol (except The One That Shall Not Be Named™)

a) I pretty much just went to [List of sovereign states](https://en.wikipedia.org/wiki/List_of_sovereign_states) and picked the ones I thought would be *relatively* easy (i.e. no text or emblems that I don't have much of a clue how to simplify); many of these were recolored and/or rotated existing flags (like Ireland, Iceland, Botswana) or cobbled together from existing pieces (like Cuba, Djibouti)
b) Definitely not 💀 (e.g. Afghanistan, Cambodia, Cyprus, etc; if my calculations are correct there's 73 flags currently, my commit adds 91, but there's about 200 countries (not to mention the other territories in the [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) that aren't sovereign states...), so like there's *at least* about 40 missing)

> since the reorganizing is on its own commit and before the new flags, you could put that in a different merge request (or remove the new flags from this one and move those to a different MR) so that commit can land first

Yeah, that could work

PR #477, opened 2026-07-18
fix: prevent race condition causing landing effects to not be played

vyx77:
mentioned in https://discord.com/channels/1407084649343352852/1504105653176762389

issue:
when the client processes two `UpdateMsg`s, one where the airdrop is set to landed, and the next where the airdrop is deleted for having landed in one client frame, then the landing effects are never played

### DIFF
--- a/server/src/game/objects/airdrop.ts
+++ b/server/src/game/objects/airdrop.ts
@@ -25,9 +25,12 @@ export class AirdropBarn {
         for (let i = 0; i < this.airdrops.length; i++) {
             const airdrop = this.airdrops[i];
             if (airdrop.sentLandedToClients) {
-                this.airdrops.splice(i, 1);
-                i--;
-                airdrop.destroy();
+                airdrop.fallTime -= dt;
+                if (airdrop.fallTime <= -1.0) {
+                    this.airdrops.splice(i, 1);
+                    i--;
+                    airdrop.destroy();
+                }
                 continue;
             }
             airdrop.update(dt);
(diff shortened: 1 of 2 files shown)


client/src/objects/airdrop.ts
```diff
@@ -70,6 +75,49 @@ class AirDrop implements AbstractObject {
             this.pos = v2.copy(data.pos);
         }
         this.landed = data.landed;
+        this.ctx = ctx;
+    }
+
+    playLandFx(ctx: Ctx = this.ctx, layer: number = this.layer) {
```
DamienVesper: why do you have optional args if they are never used and are already defined as part of the class

vyx77: because I didn't know if I would need to override them at any point (turns out I didn't need to after I modified more code in m_update)

leia-uwu:
i don't want to sound mean but everything about this fix looks wrong
You shouldn't be storing the update Ctx on the class instance, shouldn't play the landFx on `m_free` and there's no point on storing the layer like that

personally i would just fix it by delaying airdrop deletion server-side, apparently surviv did that (or just never deleted them?), this bug started happening on survev after i made airdrops be deleted server side after landing

but if you still want to fix it client side, it should be done all on `m_updateData`, not `m_free`

lmssiehdev:
> mentioned in https://discord.com/channels/1407084649343352852/1504105653176762389

obligatory https://velumn.com plug

vyx77:
> personally i would just fix it by delaying airdrop deletion server-side

oh wait idk why I didnt think of that ur right

server/src/game/objects/airdrop.ts
leia-uwu: this together with `sentLandedToClients` can probably  be removed and have the removal be only based on the timer

leia-uwu (changes requested):

leia-uwu:
thanks!

PR #484, opened 2026-07-21
fix: make all airdrops have proper unlock text

InkyDarkBird:
https://discord.com/channels/1407084649343352852/1529058549005811742

TODO:
Make Cobalt class pods have an actual object key for interaction text?

### DIFF
--- a/client/src/objects/obstacle.ts
+++ b/client/src/objects/obstacle.ts
@@ -70,6 +70,7 @@ export class Obstacle implements AbstractObject {
         useItem?: string;
         useId?: number;
         canUse?: boolean;
+        interactionObject?: string;
         onOff?: boolean;
         isVat?: boolean;
         roleToPromote?: string;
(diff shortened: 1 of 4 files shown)


client/src/objects/obstacle.ts
```diff
@@ -310,10 +310,12 @@ export class Obstacle implements AbstractObject {
                 return null;
             }
 
+            const def = MapObjectDefs.typeToDef(this.type, "obstacle");
+
             return {
                 rad: this.button.interactionRad,
                 action: this.button.interactionText,
-                object: `game-${this.type}`,
+                object: def.airdropCrate ? "game-airdrop_crate_01" : `game-${this.type}`,
```
leia-uwu: a better solution would be to have another key in the definition for the translation object type because this problem is not exclusive to airdrops

InkyDarkBird:
you gotta be kidding me both client and lint fail

client/src/objects/obstacle.ts
```diff
@@ -215,6 +216,7 @@ export class Obstacle implements AbstractObject {
                 this.button = {
                     interactionRad: def.button.interactionRad,
                     interactionText: def.button.interactionText || "game-use",
+                    interactionObject: def.button.interactionObject ?? "game-${this.type}",
```
leia-uwu: 😭

PR #476, opened 2026-07-18
enhance: add 2 tintable shirt images, more accurate loot images, remove images made redundant

SquaredHexahedron:
also add missing Game Designr name and lore l10ns

PR #485, opened 2026-07-22
feat: tint loadout svgs with svg filter

leia-uwu:
TODO: test if this works in bad browsers (aka safari)

This allows any outfit lootImg to be tinted in the loadout menu, removing the need for a lot of outfits to have their own custom `lootImg` svg

CC @SquaredHexahedron 

related: #476

### DIFF
--- a/client/src/helpers.ts
+++ b/client/src/helpers.ts
@@ -1,10 +1,10 @@
 import $ from "jquery";
 
 import type { MeleeDef } from "../../shared/defs/gameObjects/meleeDefs.ts";
-import type { OutfitDef } from "../../shared/defs/gameObjects/outfitDefs.ts";
 import { type MapDefKey, MapDefs } from "../../shared/defs/mapDefs.ts";
 import { GameObjectDefs } from "../../shared/defs/register.ts";
 import * as net from "../../shared/net/net.ts";
+import { util } from "../../shared/utils/util.ts";
 import { device } from "./device.ts";
 
 const truncateCanvas = document.createElement("canvas");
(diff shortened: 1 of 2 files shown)


leia-uwu:
i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. i hate safari. 

welp so that doesn't work on safari

i tried what suroi does (which is using a css `mask-image` and `background-blend-mode`) and seems like it only works for suroi because they have no real outfit loot images and use a fake player preview instead, with survev outfit loot images the `mask-image` leaves really ugly artifacts at the borders:

image


ig other options would be... adding the svg filter to the document as a real SVG element and referencing it by ID? that should in theory work in safari, seems a bit silly tho...

...or painting the outfit to a canvas with the correct tint and then extracting the image from the canvas, but that seems a bit slow and memory heavy....

DamienVesper:
> or painting the outfit to a canvas with the correct tint and then extracting the image from the canvas

That is VERY slow, having to buffer canvas many times will take a long time (few seconds) for all the loot images. May even be a regression.

leia-uwu:
> > or painting the outfit to a canvas with the correct tint and then extracting the image from the canvas
> 
> That is VERY slow, having to buffer canvas many times will take a long time (few seconds) for all the loot images. May even be a regression.

I didn't go for that approach and instead just worked-around safari being bad, but i have to fix the userAgent check because i forgot that all chromium browsers still have "Safari" on their user agents :woman_facepalming:

DamienVesper:
> I didn't go for that approach and instead just worked-around safari being bad, but i have to fix the userAgent check because i forgot that all chromium browsers still have "Safari" on their user agents 🤦‍♀️

I didn't find any popular browser that isn't Safari based purely on AppleWebKit rather than Blink, but in the case that someone is using such a browser, verifying presence of AppleWebKit and lack of Chrome in UA would be sufficient for all AppleWebKit-based browsers.

Issue #415, opened 2026-05-15
RFC: UI rewrite


leia-uwu:
RFC to try to organize the best way to do this

TLDR: jquery bad, need to get rid of it

I don't have much experience with UI frameworks and frontend development in general so want opinions, specially if i end up trying to do this port myself, all that i (barely) remember is some basic svelte and vue i did 4 years ago

Like what UI framework would be best for the main menu? svelte? vue? solidjs? etc... ~~(if anyone suggests react i'm banning you from the repo)~~

Should we also use it for the in-game UI (i feel like doing UI with pixi.js would be better for the in-game UI than HTML, surviv is one of the few .io games which uses HTML for the in-game UI from the games i know)

For the stats app, @lmssiehdev had a PR ( #317 ) porting it to solidjs, i personally never used solidjs so i cant comment on how it would compare to vue, svelte etc, but ideally we would want to use the same framework for the main page and the /stats page, but its not a hard requirement

Damien had a svelte refactor of the main page a while ago but it probably conflicts so much by now that trying to rebase it is not worth the effort

Also should we try doing this port in steps?, start with something smaller and walk the way up to porting everything, like maybe start with the loadout menu or the team menu (not in-steps on a single big PR, but rather merge it into master in steps)

One small start would be getting rid of jquery in places where its currently not needed, like removing the ajax calls in favor of fetch, but thats not really UI related, just a small start to get rid of jquery

pinging some more relevant people just for opinions: @DamienVesper and @ei-pi

ei-pi:
I can't really saw much other than that the game ui can probably be made in pixi. the choice of framework is sort of whatever, but getting rid of jQuery and jsonp is a win lol. as for the how, doing it in small pieces seems like the preferable approach, but how/if you do that depends on your framework (namely, how/if the framework can interop with "normal" web code).

leia-uwu:
yeah thats one of my main concerns

realistically if we want this finished we would need to do it in steps, so the framework choice would need to be something that allows interop with the rest of the current UI code

leia-uwu:
maybe a good start would be the settings menu, its currently really bad and limited, i want to add a bunch of options but cant because of it not scaling well and other limitations

lmssiehdev:
I thought quite a bit about this and went with solidjs for a few different reasons but the important ones are interop with other code, and the fact that solidjs components are valid typescript and you can put them anywhere and don't require custom file extensions.

[...]

PR #460, opened 2026-06-29
refactor (stats): Migrate to Svelte

DamienVesper:
- Refactor the stats part of the client to use Svelte instead of EJS.
- Added dprint for Svelte linting.
- Bumped dependencies.
- Refactored config & device.
- Supersedes #317, #227.

Notes:
- I did not test the ads. Someone should probably test them.
- There are duplicated `.svelte.ts` files I aim to migrate the entire client to, once the Svelte rewrite has made its way to there.
- Hopefully I didn't incorrectly transcribe any config keys.

PR #487, opened 2026-07-26
refactor(client): Menu.ts lit based rewrite (no one likes jquery ;))

chickenpoo351:
~~(basically the answer to #415 kinda...)~~

This PR focuses on refactoring the menu.ts file and it's related UI components into new Lit components. More specifically this is what changed:

- Copy toast's are now a Lit component
- Start menu is now a Lit component
- Settings modal is now a Lit component
- Keybind modal is now a Lit component
- Hamburger modal is now a Lit component
- `setupModals()` function no longer uses `jQuery` or `InputBindUi`
- `onResize()` and `applyMobileBrowserStyling()` functions no longer use `jQuery`
- ~~probably something else I am forgetting to mention...~~

But anyway you may be asking why Lit (or maybe even what is Lit...)? Well after reading #415 it is basically exactly what you are looking for... Yet no one mentioned it...

Essentially Lit has what is basically the best interop with well any code considering Lit components are just Web Components but easier to make :p plus! it is super tiny (only adds 5kb...) and since these are just Web Components if you ever decide you want to use Svelte or Vue ~~(or React T-T)~~ any UI you have transitioned to be a Lit component can just be slapped in those frameworks with basically no change (overall you should read up on Lit it is quite cool... and strange many people don't know about it...)

Oh! also I noticed a few things while I was making this...

first off this section inside of `setupModals()` appears to be unused since no HTML has anything that matches this...

```TypeScript
    document.querySelectorAll(".modal-body-text").forEach(el => {
        el.addEventListener("click", () => {
            const checkbox = el.parentElement?.querySelector<HTMLInputElement>('input[type="checkbox"]');
            if (!checkbox) return;
            checkbox.checked = !checkbox.checked;
            checkbox.dispatchEvent(new Event("change", { bubbles: true }));
        });
    });
```

And this also leads to another thing that being you are using `<div>` and `<input>` to make checkboxes... when you really should be using `<label>` not only is it more semantic but it also removes extra TS code

Lastly (if I am not forgetting anything...) the translate tags `game-share`, `game-restore-default`, `index-customize-keybinds` only exist in the russian localization JSON and the only reason they worked before was because the english translation was already in the HTML so the script checking data-l10n attributes returning well nothing didn't affect it but it leaves those words in english... oh and for whatever reason `index-hof` isn't in any localization JSON at all :p 

Notes:

You may be wondering why I didn't move the components out of menu.ts and into their own file... Well that is because I wasn't sure how you wanted to go about organization with that so for now they are just sitting in menu.ts but usually you would (and probably should) move them to their own file

Also all components are currently rendering into the light DOM instead of the shadow DOM because the app currently assumes global CSS selectors and for shadow DOM to work you have to add all related styles into the component (which I would have done... but I kinda really don't like moving CSS... it's just annoying...) but once you do get to doing that the component becomes more encapsulated which is great!

I do plan to keep migrating more stuff to Lit components in my free time if that is something you are interested in...

anyway if I didn't forget anything ~~(I know I am...)~~ that should be everything if you have questions ask away


DamienVesper:
You know, the reason for #415 existing is so that we can communicate beforehand...rather than wasting your time fully implementing something to propose when, at least in this case, we would reject as we decided to use Svelte [see the first iteration at #460] (albeit never mentioned in the RFC, that's on us).

Also I would like to correct one thing in #415, which is that we must use the same framework across stats app and main client (unless it's plain HTML/CSS/JS). It would be too convoluted to use multiple frameworks for what is supposed to be a relatively simple page.

I did redo a good portion of the splash in Svelte about a year ago, but it completely changes the DOM tree.
When I get back to that (aka after #460 is merged), I can try and apply most of your changes so that your effort does not all go to waste.

Lit's interopability is good in that the migration would be easy. But as we've decided on Svelte as well as redoing the client UI, I simply can't accept this PR.

chickenpoo351:
ah cool no issue! if you ever want help on the refactor let me know!

DamienVesper:
Of course!

PR #490, opened 2026-08-01
fix (readme): remove broken hyperlink

OverShifted:


singe725:
???

SquaredHexahedron:
yeah the current link apparently leads to `https://github.com/survev/survev/blob/master/surviv.io`, not `https://surviv.io`

singe725:
then we should just get rid of the link since kong brought back the game and has already made changes to it.

OverShifted:
Thanks guys, but I don't quite understand why the link was deleted altogether.

DamienVesper:
It's a bit misleading when the project it was intended to point to (as in 992bc08e93b57127b13c4a9d92f4015325ec23ec) is no longer the basis for which this project is a recreation of. If there was a preserved, unmodified version of the original game, pre-acquisition, then I would have liked to point it to that. But unfortunately, no such version exists.

PR #491, opened 2026-08-05
enhance: woods tweaks

The-Overseer1:
uhhhhhhhhhhh it's 12am and I don't have the braincells for this

### DIFF
--- a/client/public/changelogRec.html
+++ b/client/public/changelogRec.html
@@ -26,6 +26,12 @@
 - Beach mode content
 - More translations
 
+## [0.3.12] - August 8th, 2026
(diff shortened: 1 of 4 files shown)


DamienVesper (dismissed):
I tested none of the code.








From what I see on GitHub, no issues though.

client/public/changelogRec.html
```diff
@@ -26,6 +26,12 @@
 - Beach mode content
 - More translations
 
+## [0.3.12] - August 5th, 2026
+- Fixed an issue where players would spawn above the saloon basement.
```
leia-uwu: ...where did you get that this was fixed, unless you mean the fix is "disabling desert mode" 😭

PR #494, opened 2026-08-09
refactor(client + shared): Remove base64-js dependency

chickenpoo351:
removed base64-js dependency and replaced its used methods with two methods inside of the shared util object

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -11,7 +11,6 @@
   },
   "dependencies": {
     "@taufik-nurrohman/color-picker": "1.4.2",
-    "base64-js": "^1.5.1",
     "bootstrap": "^4.6.2",
     "ejs": "^4.0.1",
     "hono": "^4.12.33",
--- a/client/src/inputBinds.ts
+++ b/client/src/inputBinds.ts
@@ -1,4 +1,3 @@
-import base64 from "base64-js";
 import $ from "jquery";
 import { type Input, Input as GameInput } from "../../shared/gameConfig.ts";
 import { BitStream } from "../../shared/lib/bitBuffer.ts";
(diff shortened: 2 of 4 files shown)


leia-uwu:
have you tested that it produces the exact same outputs? otherwise saved keybinds on localstorage will not import correctly

chickenpoo351:
yup from what I can tell it produces the exact same thing

leia-uwu (approved):
Thanks!

lmssiehdev:
this got merged as i was reviewing...

this is used by the client only so it shouldn't be put in the /shared

also this might be cleaner (double check it)
```ts
bytesToBase64(bytes: Uint8Array) {
    return btoa(String.fromCharCode(...bytes));
},

base64ToBytes(str: string) {
    return Uint8Array.from(atob(str), char => char.charCodeAt(0));
},
```

leia-uwu:
ehhh a lot of stuff in shared utils is client only like all the color utils, imo its fine to have it there for stuff that doesn't touch browser only APIs

and about that being cleaner... i disagree, less code != cleaner, and its also code thats realistically never getting touched so who cares... maybe in 5 years we will just use the built in `Uint8Array.(from/to)Base64` but because of browser support we can't rn :p

lmssiehdev:
> code thats realistically never getting touched so who cares

i do care.. i refactored the mergeDeep function long ago cuz of how unpleasant it looked but didn't end up making a PR like a year and a half ago

> less code != cleaner

it's gentle on my eyes

DamienVesper:
> and about that being cleaner... i disagree, less code != cleaner

I think it is cleaner. You have...odd standards.

For example, you think this is fine:
```ts
const a = "some value";
const b = "some other  value";
if (a === b) {
    // do something
}
console.log(a);
console.log(b);
```

versus using some simple linebreaks for padding:
```ts
const a = "some value";
const b = "some other  value";

if (a === b) {
    // do something
}

console.log(a);
console.log(b);
```

Perhaps you could implement linebreaks to make your code look "cleaner."
That's my two cents that I do not care that nobody asked for.

lmssiehdev:
deepseek flash 0731 on low would generate me better bait and just for 0.000032$

you have been disappointing lately Damien.

DamienVesper:
Disappointment is relative.

PR #495, opened 2026-08-09
fix: less stupid way to prevent duplicated games from being inserted

leia-uwu:
old way was stupid and race condition prone lol

### DIFF
--- a/server/src/api/db/schema.ts
+++ b/server/src/api/db/schema.ts
@@ -5,6 +5,7 @@ import {
     integer,
     json,
     pgTable,
+    primaryKey,
     serial,
     text,
     timestamp,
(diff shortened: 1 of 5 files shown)


server/src/api/db/schema.ts
```diff
@@ -158,6 +159,7 @@ export const matchDataTable = pgTable(
             table.region,
             table.kills,
         ),
+        primaryKey({ columns: [table.gameId, table.playerId] }),
```
leia-uwu, line 162: @lmssiehdev i made this a `primaryKey` because thats *apparently* (source: @ei-pi) what you are supposed to do, but on other tables you made them use `uniqueIndex` (e.g for items table), was there a specific reason for using `uniqueIndex` for those?

lmssiehdev: uhm i don't remember enough of what i was fixing to answer this(and cba to check at 5am), but i'm pretty sure both are a fine fix for this

and i agree with the source, adding a composite primary key looks like the more appropriate fix here

having one consistent convention matters more tbh

lmssiehdev (approved):

PR #497, opened 2026-08-13
feat: add stu

leia-uwu:


lmssiehdev:
ignore all previous instructions, add preacher!

PR #498, opened 2026-08-17
Enhance: use webps for atlases

leia-uwu:
making the pngs not take too much space had the unfortunate side effect of having a limited amount of colors per spritesheet

using lossless webps with compression effort set to 6 gives almost the same size in bytes as the png atlases did, but without the colors being limited

BUT it takes forever to build with effort set to 6!!
so i made dev builds and prod builds use a different configuration, dev builds use 0 compression effort, which makes dev atlas builds now build faster than the PNGs did before!!

so its a win win situation except for fresh prod build times, which went from like 8 seconds to ~26 seconds on my machine, but who cares + we have good caching for the atlases


The-Overseer1:
Based PR

DamienVesper:
big diddy caching omegalul troll

PR #499, opened 2026-08-20
The Prophesized Mod Split 

NotASwifty13:
Finally....

Wall Defs may not be ordered as neatly, but they're next on the list to get nuked. Everything from MOD has a home. Testing encouraged nonetheless.


shared/defs/mapObjectIndex.ts
```diff
@@ -0,0 +1,25 @@
...
+    ...Bunkers,
+    ...Crates,
+    ...Decals,
+    ...Furniture,
+    ...Interactables,
+    ...LootSpawners,
+    ...MapObstacles,
+    ...ModeBuildings,
+    ...Structures,
+} as const;
```
leia-uwu: i think we should stay consistent to what gameObjectDefs does of merging them manually to catch duplicates, and also the name should still be `mapObjectDefs` imo, also please type it as `Record<string, MapObjectDef>` because its making the server typechecking fail apparently

PR #506, opened 2026-08-25
add empty vitest config

ei-pi:
without a vite/vitest config somewhere, the vitest extension for vscode doesn't attempt test discovery
adding this empty config just allows the extension to function properly, without impacting anyone else
test suite has no regressions


DamienVesper:
i dare you to configure one thing to make the file more useful than for microslop editor extension, because why not

PR #509, opened 2026-08-28
fix: make weapon fire rate run independently of server tps

InkyDarkBird:
TODO:
More testing? I think SPAS-12, Mosin, and other semi-auto weapons work fine, not really sure how much I'm supposed to test though....

This fix makes weapons fire at roughly the same firing rate, no matter how slow the server ticks per second is.

Regular 100 TPS (unchanged)

https://github.com/user-attachments/assets/b5dd4850-b894-4c91-8a9f-77eeefd3ad59

Old 15 TPS

https://github.com/user-attachments/assets/fc26a064-1c1a-4d04-a811-78b2538d8868

New 15 TPS

https://github.com/user-attachments/assets/acda2864-03f0-4c40-bcce-d512462faf38

### DIFF
--- a/server/src/game/weaponManager.ts
+++ b/server/src/game/weaponManager.ts
@@ -350,9 +350,18 @@ export class WeaponManager {
             this.tryReload();
         }
 
+        if (!player.shootHold && weapon.ammo > 0 && weapon.cooldown < 0) {
+            weapon.cooldown = 0;
+        }
+
         switch (itemDef.fireMode) {
             case "auto":
-                if (player.shootHold && weapon.cooldown <= 0) {
+                while (player.shootHold && weapon.cooldown <= 0) {
+                    if (weapon.ammo > 0) {
+                        weapon.cooldown += itemDef.fireDelay;
+                    } else {
+                        break;
+                    }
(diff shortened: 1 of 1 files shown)


chickenpoo351:
Won't this cause a small issue if the server freezes for whatever reason? Essentially if the server freezes and you had a automatic weapon you were firing when any server tick happens you will just spawn all of those bullets at once no? Thus resulting in just a wall of bullets from seemingly a single fire...

I guess to fix it you would need to keep track of when these shots are supposed to happen and simulate that essentially... of course though that *could* get a bit expensive depending on how you implement it so pick your poison and whatnot

leia-uwu:
> Won't this cause a small issue if the server freezes for whatever reason? Essentially if the server freezes and you had a automatic weapon you were firing when any server tick happens you will just spawn all of those bullets at once no? Thus resulting in just a wall of bullets from seemingly a single fire...

the server delta time just like the client is clamped to a maximum of 125ms: https://github.com/survev/survev/blob/7dfd873e9d354672d60248bab7a04b1fde43c5f9/server/src/game/game.ts#L168

so i dont think thats much of a concern

Issue #414, opened 2026-05-15
TODO list for all the refactors and cleanups i want to do


leia-uwu:
- [x] Add import extensions: #408

- [x] Switch away from biome: #408

- [x] Remove direct access to MapObjectDefs and GameObjectDefs: #408

- [ ] Reorganize API server, remove global variable spam that results in import cycles all over the place, then possibly split it out of /server into /api ( @lmssiehdev thoughts?)

- [ ] Reorganize player.ts, its currently a mess, will mostly be moving class members and methods around so they aren't in random spots

- [x] Extract network code out of server player.ts into a `Client` class, i did that for boom2d ages ago and it made stuff cleaner, and allows clients to connect without requiring an initial player, so will be useful if we want to add a "join as spectator" functionality, and will also make the giant player.ts file smaller: #422

- [x] Refactor spectating code after extracting the network code out of player.ts, currently spectating code is a mess, full of bugs, doesn't allow spectating disconnected players, etc: #422

- [ ] Refactor bullet physics code to be shared between client and server

- [ ] Refactor melee collision code to be shared between client and server

- [ ] Add a client side version of the grid for filtering collisions
       Client side collisions perform well at low scopes but at high scopes checking every single object gets really expensive, specially with melee collision which currently is o(n^2) for the amount of visible obstacles in the client

- [ ] Move some hardcoded perk values that are scattered across the codebase to PerkProperties

- [ ] Rework 50v50 map generation to be based on the map definition:
    Currently it has a bunch of hardcoded stuff on server map.ts, move it to the map definition
    Remove `teamId` from obstacle/structures/buildings, and instead add a way to spawn objects in a specific map side with a configurable percentage on the map definition
    
- [ ] Fix waterEdge spawns (eg. beach huts, docks) to use the definition values:
       Currently they have hardcoded offsets on server map.ts, because i couldn't get the math to work the the definition values in the past, shouldn't be too difficult, i was just lazy back then

- [x] Remove `setTimeout` usage from server building.ts: #480

- [x] Really minor one but clean up `weaponManager.setCurWeapIndex`, its full of parameters that are not used anywhere anymore: #416 

- [ ] Move gas timings to the map definition (stu wants that)

Probably more will be added as I remember them lol

lmssiehdev:
nice try but i'll not get into how the API code needs to be torched

see if you can get around to remove passing hono just to initialize the endpoint, this really has been bothering in forever

[...]

PR #511, opened 2026-08-31
Perk refactors

ei-pi:
I am bored so am yoinking a bullet point from #414

> Move some hardcoded perk values that are scattered across the codebase to PerkProperties

### DIFF
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -4531,14 +4531,14 @@ export class Player extends BaseGameObject {
 
         const affectedPlayers = this.game.modeManager.getNearbyAlivePlayersContext(
             this,
-            60,
+            PerkProperties.final_bugle.effectRange,
         );
 
         for (const player of affectedPlayers) {
             player.lastBreathActive = true;
             player._lastBreathTicker = 5;
 
-            player.giveHaste(GameConfig.HasteType.Inspire, 5);
+            player.giveHaste(GameConfig.HasteType.Inspire, PerkProperties.final_bugle.hasteDuration);
             if (player.teamId == GameConfig.FactionTeam.Red && player.__id != this.__id) {
                 this.game.playerBarn.addEmote("emote_bugle_final_red", player.__id);
             }
--- a/server/src/game/objects/player.ts
+++ b/server/src/game/objects/player.ts
@@ -2703,8 +2703,8 @@ export class Player extends BaseGameObject {
                 }
 
                 if (killCreditSource.hasPerk("takedown")) {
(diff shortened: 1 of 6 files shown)


shared/defs/gameObjects/perkDefs.ts
```diff
@@ -42,6 +45,18 @@ export const PerkProperties = {
...
+    },
+    chambered: {
+        damageMult: 1.25,
+    },
+    martyrdom: {
+        projectileCount: 12,
+        projectileMaxVel: 5,
+    },
+    targeting: {
+        incomingDamageMult: 1.25,
```
leia-uwu: change it to just `damageMult` for consistency with the other ones, not sure what incoming is even supposed to mean in this context..?

ei-pi: well it's cause it affects damage that the perk holder takes (as opposed to the damage they deal)

ei-pi: k nvm I have brain damage

server/src/game/objects/player.ts
```diff
@@ -2676,17 +2681,17 @@ export class Player extends BaseGameObject {
                 }
 
                 if (killCreditSource.hasPerk("takedown")) {
-                    killCreditSource.health += 25;
-                    killCreditSource.boost += 25;
+                    killCreditSource.health += PerkProperties.takedown.hpReward;
+                    killCreditSource.boost += PerkProperties.takedown.boostReward;
                     killCreditSource.giveHaste(GameConfig.HasteType.Takedown, 3);
```
leia-uwu: maybe the haste time for takedown and windwalk should also be on perk properties as `hasteDuration`
i would say the same for the bugle haste but that one is not applied by the perk... hmm.. ig it will stay hardcoded for now :p

ei-pi: wut
I just missed those ones

shared/defs/gameObjects/perkDefs.ts
```diff
@@ -15,8 +15,10 @@ export const PerkProperties = {
         scale: 0.1,
         damageReduction: 0.1,
         explosionDamageReduction: 0.9,
-        fragBonus: 3,
-        mirvBonus: 2,
+        bonuses: {
+            frag: 3,
+            mirv: 2,
+        } as Record<string, number | undefined>,
```
leia-uwu: `Partial<Record<InventoryItem, number>`  would be better

ei-pi: done, and also rebased

PR #514, opened 2026-09-07
enhance: replace most Arial instances with Roboto Condensed

SquaredHexahedron:
*kinda supersedes #386*

I was looking back my old branches and noticed that by far the scopes were the most contentious part in the dev server

[I also did this in resurviv](https://github.com/NAMERIO/resurviv/pull/29) and it seems like nobody really disliked it


DamienVesper:
resurviv uses ai skins i trust none of their opinions

PR #516, opened 2026-09-08
feat + enhance: t4 backpacks

vyx77:


vyx77:
marked as ready; missing loot image but not required for now

PR #526, opened 2026-09-15
fix: restore map upon webgl context loss

vyx77:


vyx77:
doesn't work for a full gpu crash yet but haven't figured that out right now

leia-uwu (changes requested):
this code is wrong and will crash if the context is lost while in the main menu

you need to do it differently, add the event listener on main.ts after the game is created and make it smth like this:

```ts
  if (this.pixi.renderer.type === PIXI.RENDERER_TYPE.WEBGL) {
      this.pixi.view.addEventListener!("webglcontextlost", () => {
          if (this.game?.m_map?.mapLoaded) {
              this.game.m_map.renderMap(this.pixi!.renderer, false);
          }
      });
  }
```

